PR #49, opened 2015-01-02
Preperations for version 2.x

excpt:
I think it is time to clean up the single module behavior of the jwt gem to keep up with the specs of JWT and the underlying specs (JWS, JWE, JWA, JWK). As a result the gem should match at least the JWT specifications for the signing of JSON Web Tokens. Encryption may be another goal to but I think it shouldn't be the main approach.

The code is not complete yet. I'd like to get feedback or a code review for the current code base and maybe we can add a milestone and figure out what should be implemented in the version 2.x branch.

What I've done so far:

* Refactoring the code base, including tests
* removed validators - should be moved into a seperate validation module/class
* Split code into modules which manage each section of the specs (JWT and JWA)
* removed the Ruby 1.8 support (but should return for old projects that cannot switch to a later and supported version of Ruby)

Goals:

* JWT interface should stay the same
* Implement the missing algorithms for signing specified in the JWA specs
* Implement JWS
* Implement JWT validators
* Add more tests, more different use cases


README.md
```diff
@@ -88,21 +125,27 @@ For example, if you have a JWT payload with a expiration time set to 30 seconds
 after creation but you know that sometimes you will process it after 30 seconds,
 you can set a leeway of 10 seconds in order to have some margin:
 
-    jwt_payload = JWT.encode({'exp': Time.now.to_i + 30}, 'secret')
-    sleep(32)
-    # jwt_payload is now expired
-    # But with some leeway, it will still validate
-    JWT.decode(jwt_payload, 'secret', true, leeway=10)
+```ruby
+jwt_payload = JWT.encode({'exp': Time.now.to_i + 30}, 'secret')
```
robotvert: Maybe I'm just nitpicking here but in the code blocks throughout the README you're using different hash syntaxes, double and single quotes etc.
Since Ruby 1.8 is removed in that PR how about replacing all of these with the short notation?
(e.g. `{ exp: Time.now.to_i }`)

excpt: That's right. I will change this and update the PR.

excpt: Done. :)

PR #56, opened 2015-02-20
Added support for not before claim verification

punkle:
Added support for not before claim verification according to http://self-issued.info/docs/draft-ietf-oauth-json-web-token.html#nbfDef


excpt:
Thanks for the pull request! :tada: 

All tests pass. :white_check_mark: 

@progrium You may release a new gem version now. :)

PR #60, opened 2015-03-03
Clarify .encode API in README

jbodah:
I lost a lot of time on this. For some reason I was thinking the syntax was `{ header => payload }` from the examples


excpt:
Tanks for the documentation improvement. :)

PR #65, opened 2015-03-10
When throwing errors, mention expected/received values

rolodato:
Also changed strings to single-quote when not interpolating.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
-source "https://rubygems.org"
+source 'https://rubygems.org'
 
 gem 'json', '>= 1.2.4'
 gem 'multi_json', '~> 1.0', :platforms => :ruby_18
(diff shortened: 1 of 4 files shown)


excpt:
:+1: Thanks :-)

ZhangHanDong:
Why must change single-quote?   
http://viget.com/extend/just-use-double-quoted-ruby-strings

excpt:
@ZhangHanDong It's just a code convention or best practice. It's not about speed or anything. It's common to use single quotes for all strings unless you have to pass some values to the string then you use double quotes "Hi #{name}".

ZhangHanDong:
@excpt  I see,  just  i like use double quoted .  all right.   : )

rolodato:
Yes, it's the convention and it's clearer to see at a first glance if something's being replaced or it's a constant string. Also, some IDEs like RubyMine mark it as a warning by default (double-quote string without interpolation).

PR #71, opened 2015-04-01
Add an option to verify the signature on decode

javawizard:


danleyden:
Looks like a good idea. I wonder if it would be a good idea for the 'allowed algorithm' to be an array? In many cases, I would expect multiple algorithms to be acceptable rather than requiring a single exact match.

excpt:
@javawizard Thanks for the PR. I will check and merge it tonight.

@danleyden IMO checking only one algo is the better solution. Checking against different algorithms is a feature that depends on each individual project. We should simply focus on the stability of the lib and keep things simple. It may be a feature for version 2.x to handle multiple allowed algos that the ruby-jwt will check against.

AlexParamonov:
Hello, when it will be pushed to Rubygems? We need to specify the algorithm to avoid vulnerabilities.

PR #72, opened 2015-04-03
Fixed some examples to make them copy-pastable

jer:


excpt:
Thank you.

Version 1.x still support Ruby 1.8 and 1.9 which doesn't support the Ruby 2.x syntax.

Merged! :beers:

PR #85, opened 2015-06-10
Code review: Rspec test refactoring

excpt:
Code review request. Please comment! :)


lib/jwt.rb
```diff
@@ -163,17 +161,17 @@ def decode(jwt, key=nil, verify=true, options={}, &keyfinder)
     if options[:verify_not_before] && payload.include?('nbf')
       raise JWT::ImmatureSignature.new('Signature nbf has not been reached') unless payload['nbf'].to_i < (Time.now.to_i + options[:leeway])
     end
-    if options[:verify_iss] && payload.include?('iss')
-      raise JWT::InvalidIssuerError.new("Invalid issuer. Expected #{options['iss']}, received #{payload['iss']}") unless payload['iss'].to_s == options['iss'].to_s
+    if options[:verify_iss] && options['iss']
```
danleyden: this looks like it could lead to problems with implementers:

if the caller passes :verify_iss but not 'iss' (or the other way around), then this could leas to an unexpected successful verification.

Probably worth adding a check and failing with an argument error in the case that one is specified but not both.

lib/jwt.rb
```diff
@@ -163,17 +161,17 @@ def decode(jwt, key=nil, verify=true, options={}, &keyfinder)
...
+      raise JWT::InvalidIssuerError.new("Invalid issuer. Expected #{options['iss']}, received #{payload['iss'] || '<none>'}") unless payload['iss'].to_s == options['iss'].to_s
     end
     if options[:verify_iat] && payload.include?('iat')
       raise JWT::InvalidIatError.new('Invalid iat') unless (payload['iat'].is_a?(Integer) and payload['iat'].to_i <= Time.now.to_i)
     end
-    if options[:verify_aud] && payload.include?('aud')
+    if options[:verify_aud] && options['aud']
```
danleyden: same concern as above - providing one but not both should be an argument error

lib/jwt.rb
excpt: Thanks for the review :)

Going to add more tests and implement checks for these cases.

spec/jwt_spec.rb
```diff
@@ -1,434 +1,502 @@
...
 describe JWT do
-  before do
-    @payload = {'foo' => 'bar', 'exp' => Time.now.to_i + 1, 'nbf' => Time.now.to_i - 1 }
+  let :payload do
+    {
+      user_id: 'some@user.tld'
+    }
+  end
+
+  let :data do
```
danleyden: given that this isn't overridden, it might be worth making this a constant instead of a let block.

as-is, each of those files will be re-read for any it block that accesses data

spec/jwt_spec.rb
```diff
@@ -1,434 +1,502 @@
...
   end
 
-  it 'encodes and decodes JWTs' do
-    secret = 'secret'
-    jwt = JWT.encode(@payload, secret)
-    decoded_payload = JWT.decode(jwt, secret)
-    expect(decoded_payload).to include(@payload)
-  end
+  context 'alg: NONE' do
+    let :alg do
```
danleyden: stylistic note (would be good to get some thoughts)...

typically, for a let block this short, I would write it as a one-liner:

```let(:alg) { 'none' }```

not sure which is preferred here

File: spec/jwt_spec.rb [hunk 5]
```diff
@@ -1,434 +1,502 @@
...
-      expect(header['kid']).to eq('default')
-      private_key.public_key
+  %w(HS256 HS384 HS512).each do |alg|
+    context "alg: #{alg}" do
+      it 'should generate a valid token' do
+        token = JWT.encode payload, data[:secret], alg
+
+        expect(token).to eq data[alg]
+      end
+
```
danleyden: this might be a good case for more judicious use of let and subject blocks, along with custom matchers (albeit, it could add more complexity so maybe not...) to end up with something like:

```ruby
RSpec::Matchers.define(:have_payload) { |expected| match { |actual| actual[1] == expected } }
RSpec::Matchers.define(:have_header) do |key, value|
  match { |actual| actual[1].key?(key) && actual[1][key] == value }
end

let(:secret) { data[:secret] }
let(:decoded_token) { JWT.decode(data[alg], secret) }
let(:decoded_payload) { decoded_token[0] }
let(:header) { decoded_token[1] }

subject { decoded_token }

describe 'a valid token' do
  it { is_expected.to have_payload(payload) }
  it { is_expected.to have_header('alg', alg) }
end

describe 'a token with the wrong secret' do
  let(:secret) { 'wrong_secret' }
  subject { -> { decoded_token } }
  it { is_expected.to raise_error(JWT::DecodeError) }

  context 'when verify is set to false' do
    let(:decoded_token) { JWT.decode(data[alg], secret, false) }
    it { is_expected.not_to raise_error }
    describe 'the decoded token' do
      subject { super().call }
      it { is_expected.to have_payload(payload) }
      it { is_expected.to have_header('alg', alg) }
    end
  end
end
```

This would have the benefit that a lot of the plumbing could be reused elsewhere in this file and (hopefully) improve readability throughout.

Of course, that style isn't something that everybody likes...

spec/jwt_spec.rb
```diff
@@ -1,434 +1,502 @@
...
+      it 'should decode a valid token' do
+        payload, header = JWT.decode data[alg], data[:secret]
+
+        expect(header['alg']).to eq alg
+        expect(payload).to eq payload
```
danleyden: payload will always eq payload... comparing the same objects.

I think you meant to compare the payload from the let block with the payload declared a couple of lines above, but because they have the same name, badness will ensue.

This looks to be repeated further down in the code too

spec/jwt_spec.rb
```diff
@@ -1,434 +1,502 @@
...
-    example_payload2 = {'hello' => 'world', 'aud' => ['url:pnd', 'aud:yes']}
-    example_secret = 'secret'
-    example_jwt = 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJoZWxsbyI6IndvcmxkIiwiYXVkIjoidXJsOnBuZCJ9._gT5veUtNiZD7wLEC6Gd0-nkQV3cl1z8G0zXq8qcd-8'
-    example_jwt2 = 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJoZWxsbyI6IndvcmxkIiwiYXVkIjpbInVybDpwbmQiLCJhdWQ6eWVzIl19.qNPNcT4X9B5uI91rIwbW2bIPTsp8wbRYW3jkZkrmqbQ'
-    decoded_payload = JWT.decode(example_jwt, example_secret, true, {'aud' => 'url:pnd'})
-    decoded_payload2 = JWT.decode(example_jwt2, example_secret, true, {'aud' => 'url:pnd'})
-    expect(decoded_payload).to include(example_payload)
-    expect(decoded_payload2).to include(example_payload2)
-  end
+      it 'should decode a valid token' do
```
aj-michael: `it should decode a valid token` and `it should generate a valid token` are identical tests.

spec/jwt_spec.rb
```diff
@@ -1,434 +1,502 @@
...
+      end
+
+      it 'should decode a valid token' do
+        payload, header = JWT.decode data[alg], data[:secret]
+
+        expect(header['alg']).to eq alg
+        expect(payload).to eq payload
+      end
+
+      it 'wrong secret should raise JWT::DecodeError' do
```
aj-michael: Personal preference: this seems like a good place for a context block.

aj-michael:
Nitpick: can you choose a consistent syntax for hash literals with symbols? grep is counting 23 occurrences of `:symbol => value` syntax and 3 occurrences of `symbol: value` syntax.

excpt:
@aj-michael This happens when I code on the train and look out of the window. Will fix that in upcoming commits.

spec/jwt_spec.rb
excpt: @aj-michael How would you merge these two tests into a single test?

spec/jwt_spec.rb
aj-michael: +1 for `let(:alg) { 'none' }`

excpt: @danleyden Some :let blocks contain a few lines. I decided to stay in the same style of code. But it saves a few lines in the specs. I am going to change the ones that can be single liners.

spec/jwt_spec.rb
aj-michael: My apologies, I'm not saying they should be identical tests. I'm saying that at the moment, they are identical, as in they contain the same expressions. I don't believe they should be.

PR #88, opened 2015-06-30
nbf check allows exact time matches.

aj-michael:
From the spec "The nbf (not before) claim identifies the time before which the JWT MUST NOT be accepted for processing.". The exact time is not before itself, so it should be valid.

This was annoying me in unit tests in a project of mine.


aj-michael:
Adding another snippet from http://self-issued.info/docs/draft-jones-json-web-token-06.html

Section 4.1 Reserved Claim Names
"The processing of the nbf claim requires that the current date/time MUST be after or equal to the not-before date/time listed in the nbf claim."

excpt:
Thanks for the commit! :+1:

PR #94, opened 2015-08-04
Fix iat claim example

larrylv:
The implementation doesn't use `options['iat']` to verify iat claim.


aj-michael:
Hi Larry, thanks for the pull request! You're certainly right, `:verify_iat` checks against the current time so there is no need for `options['iat']`.

larrylv:
@aj-michael You are welcome. :-)

Issue #102, opened 2015-09-16
aud verifies if aud is passed in, :sub does not


skippy:
If you pass in :aud and :verify_aud, then it will fail if :aud is not in the payload.

the equivalent is not true for :sub.  If you pass in :sub but it is not in the payload, it won't do anything.  I would expect the behavior of :aud, which is if you ask to verify_sub, it will fail if it isn't in the payload.

if that sounds reasonable, I'll submit a patch.  thanks!

excpt:
It should behave like the :aud validation.

PR #103, opened 2015-09-17
make sure :sub check behaves like :aud check

skippy:
if :verify_sub is turned on and options includes a :sub, make sure we
fail if the payload is not set or doesn't match.  Previously if the
payload did not have 'sub' set, it would not verify, even if verify_sub
was true

see #102


skippy:
I was tempted to also add a check where if `:verify_sub` is set but options[:sub] is not set, that an error would be thrown.  That may change the behavior a bit too much, esp. for the scope of this bug.  

But in general, if :verify is set, I would expect it to fail if `:sub` or `:aud` is not passed in.  If you would like me to, I can open another issue to track that edge case.

thanks!

PR #109, opened 2015-10-09
Fix incorrect `iat` examples

kjwierenga:
In the JWT ID Claim example in README.md the `iat` is passed in the options to `JWT.decode`. This confused me when reading the doc. The `iat` is not used in options to `JWT.decode`.

This PR removed `iat` from this example and also from the relevant specs.


review-ninja:
<a href="https://app.review.ninja/jwt/ruby-jwt/pull/109" target="_blank">ReviewNinja

excpt:
Thanks for the correction in the specs and the documentation. :beers:

PR #114, opened 2015-11-16
Updated readme for iss/aud options

ryanmcilmoyl:
When verifying issuer or audience claims, needs to pass the iss or aud keys in the options hash as symbols, not strings.  Updated the Readme to reflect this.


excpt:
Thanks! Merged.

PR #117, opened 2015-12-22
Refactor decode and verify functionality

excpt:
Move decode token logic into a separate class.
Move token verify logic into a separate module that is called by the
decode class.


review-ninja:
<a href="https://app.review.ninja/jwt/ruby-jwt/pull/117" target="_blank">ReviewNinja

PR #125, opened 2016-01-20
add RSpec task to Rakefile

jabbrwcky:
`rake test` now runs the RSpec tests as advertised in the documentation.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,1 +1,11 @@
 require 'bundler/gem_tasks'
+
+begin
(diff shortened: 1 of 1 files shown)


excpt:
HI @jabbrwcky,

thanks for the pull request.

Can you please fix following issues in you PR?

Use ruby 1.9/2.x hash style syntax:

```ruby
task default: :test
```

Display a helpful error message when you catch the exception that rspec is not present. The user may run `bundle install` in order to fix this issue.

Issue #110, opened 2015-10-09
Suggest passing block to JWT.decode for claim verification


kjwierenga:
The current method for verifying the JTI field in the token payload is only useful when the JTI value to compare against is known a priori (when making the call to `JWT.decode`).

**Summary: Using a block parameter or callback on `JWT.decode` to verify claims (passing the decoded payload/claims) would be more flexible and nicer API IMHO**

When the JTI value is dependent on information in the token payload (for example a user object found using a user_id field in the payload), then we can't pass the expected JTI value in the options hash to `JWT.decode`. In this case we don't know the expected value until after the token has been decoded. The same goes for comparing a JTI value to a blacklist.

The usefulness of the options `{verify_jti: true, jti: 'expected-value'}` passed to `JWT.decode` is limited. Maybe the same goes for `iss`, `aud` and `sub`.

_May I suggest an alternative API for verification of `jti`, `iss`, `aud` and `sub` which is to pass an (anonymous) block to `JWT.decode` for verification. The `payload` would be passed as parameters to the block. When the block returns falsey, then `JWT.decode` would raise a `JWT::VerificationError`._

For example: Given the following verification function ...

``` ruby
def valid_token?(payload)
  valid_user?(payload['user']['id']) && !blacklisted?(payload['jti'])
end
```

... with the current API I have to do:

``` ruby
payload, header = JWT.decode(token, 'my-jwt-secret')
valid_token?(payload) || raise AuthenticationError # application specific exception
```

... With a verification block this would become:

``` ruby
JWT.decode(token, 'my-jwt-secret') do |payload|
  valid_token?(payload)
end # when valid_token? return falsey raises JWT::VerificationError (for example)
```

The current verfication of the `exp`, `nbf` and `iat` claims is unambiguous so it can be handled in `JWT.decode` with the current _flag-and-expected-value-in-options-hash_ method. But I would prefer the block method to be the only method to verfiy claims. Special 'build-in' methods could be used for verification of `exp`, `nbf` and `iat` claims.

``` ruby
JWT.decode(token, 'my-jwt-secret') do |payload, header|
  valid_exp?(payload['exp']) && \ # built-in
  valid_nbf?(payload['nbf']) && \ # built-in
  valid_iat?(payload['iat']) && \ # built-in
  valid_token?(payload) # user defined
end # when valid_token? return falsey raises JWT::ValidationError (for example)
```

This method doesn't require a separate flags to indicate which claims to verify.

[...]

PR #126, opened 2016-02-09
Allow a proc to be passed for JTI verification

yahooguntu:
Since verifying a JTI isn't part of the JWT specification, I added the ability to pass a proc to JWT.decode to determine if a JTI is valid. I also updated the readme to explain how this works and to explain how the current method of validation works.

See #110


excpt:
Thanks! :beers:

Issue #153, opened 2016-06-08
[security] Signature verified after expiration/sub/iss checks


koenbollen:
Given a feature were you login using an JWT and get redirected to an URL that is inside the claims of the JWT, when an JWT is expired you are still redirected without logging in.

Because the expiration is checked before the validity of the signature an `JWT::ExpiredSignature` is raise before the signature is checked. This allows attackers to create a malicious token, with a bogus signature, that is expired but includes a redirect claim making an open redirect.

You could fix it by doing an decode with `verify_expiration: false` first, before decoding it again with all verifies on true.

Off course, you shouldn't add a feature on expired tokens but I would expect that the validity is checked before the checking the claims. Furthermore, this is just on example where it could go wrong.

It's about this code: https://git.io/votyr
I propose switching it around, verifying the validity of the signature first, and then the claims.

Thanks for the great Gem!

Cheers,
—Koen

blendle-hubbit:
Heya!

This issue was just mentioned in the Trello card "[(5) Refactor magic login two methods to one](https://trello.com/c/cqI2y7Aj)" on the [Team Buttons](https://trello.com/b/p6vfusdg) board.

Might be relevant to you, or not...

*Happy Hackin'!*

excpt:
@koenbollen can you please test your case with the #160 PR?

koenbollen:
@excpt Great work! It has fixed the issue, I've used that branch and removed my extra check and got the expected result.
Thanks.

excpt:
Thanks for the fast response. 👍

PR #160, opened 2016-08-22
Signature validation before claim verification

excpt:
The JWT signature check should always be executed before any claim verification is triggered.

As explained in #153.


koenbollen:
Code looks good to me.
:+1:

Issue #157, opened 2016-07-20
expiration check does not give "Signature has expired" error for the exact time of expiration


vnc-zestfinance:
If for example, you want to have a JWT expire at 4 hours after current time, it still doesn't expire at the exact time specified (exp field).  But it does expire like the second afterwards but according to JWT documentation, it should expire at or after the given time.  Please see the logic difference between the following two diffs:

https://github.com/jwt/ruby-jwt/blob/jwt-1.5.2/lib/jwt.rb#L161 (working version)
https://github.com/jwt/ruby-jwt/blob/v1.5.4/lib/jwt/verify.rb#L38 (not working version)

Essentially you are missing a less-than-or-equal to, I believe.  Please check it out.

PR #161, opened 2016-08-22
Fix: exp claim check

excpt:
A exp claim can only be valid if it is greater than the current time
minus the leeway. Otherwise the verification has to fail.

Details of this issue explained in  #157 by @vnc-zestfinance.


vnc-zestfinance:
looking good to me

PR #164, opened 2016-09-16
Improve error message for exp claim in payload

### LINKED ISSUE #148: JWT.decode always raises JWT::ExpiredSignature for tokens created with Time objects passed as the `exp` parameter
```
require "jwt"

secret = SecureRandom.hex(64)
exp_time = Time.now + 3600 # one hour from now
payload = {data: "test", exp: exp_time}
token = JWT.encode(payload, secret, "HS256")
JWT.decode(token, secret, true, { algorithm: "HS256" }) # => JWT::ExpiredSignature: Signature has expired
```

If I paste the generated token into [jwt.io](https://jwt.io/), this is the payload that `JWT.encode` generated:

```json
{
  "data": "test",
  "exp": "2016-04-30 13:57:44 -0500"
}
```

If I instead change the code to do `exp_time = (Time.now + 3600).to_i` (i.e. cast to Unix epoch), it works correctly and doesn't raise a `JWT::ExpiredSignature`. The payload generated by that code is obviously different:

```json
{
  "data": "test",
  "exp": 1462042886
}
```

So I'm guessing that the decoding is not converting time values like `2016-04-30 13:57:44 -0500` back into Time objects correctly.

Although the examples in the README explicitly do `.to_i`, it would be nice to either handle other time values or raise an exception if they are not supported at the encoding step.

excpt:
A better error message is displayed if a Time object is used for the exp
claim instead of an integer unix time stamp.

Fixes #148.

### DIFF
--- a/lib/jwt.rb
+++ b/lib/jwt.rb
@@ -73,6 +73,7 @@ module JWT
   end
 
   def encoded_payload(payload)
+    raise InvalidPayload, "exp claim must be an integer" if payload['exp'] && payload['exp'].is_a?(Time)
     base64url_encode(encode_json(payload))
   end
 
--- a/lib/jwt/error.rb
+++ b/lib/jwt/error.rb
@@ -9,4 +9,5 @@ module JWT
   class InvalidAudError < DecodeError; end
   class InvalidSubError < DecodeError; end
   class InvalidJtiError < DecodeError; end
+  class InvalidPayload < DecodeError; end
 end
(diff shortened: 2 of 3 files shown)


mgwidmann:
This should have never been ported back to `1.5.5` as it is not backwards compatible since the `payload` object is not guaranteed in any other location to be a `Hash` object.

Passing in `[]` or any other object which does not respond to `payload['exp']` crashes with:
```
TypeError: no implicit conversion of String into Integer
```

i.e. this crashes:

```ruby
[1] pry(main)> token = JWT.encode ['my-data', {some: 'payload'}], nil, 'none'
TypeError: no implicit conversion of String into Integer
from (pry):2:in `[]'
```

PR #172, opened 2016-10-17
Pass payload to keyfinder

CodeMonkeySteve:
I have an app that I would like to authenticate to using tokens signed with HMAC (created by the app) or ECDSA (created by other companies), depending on the value of the issuer (`iss`) claim.  I'm using the `keyfinder` block to implement that logic, but it needs to have the `payload` passed to it in order to find the appropriate public key for the issuer.  To that end, I made a small change to pass both the `header` and `payload` to `keyfinder`, if it has arity of 2.  Now my calling code looks like:

```ruby
payload = JWT.decode(token, nil, true) do |header, payload|
  alg = header['alg']
  if alg.starts_with?('HS')
    # HMAC token
    Rails.application.secrets[:secret_key_base]

  elsif alg.starts_with?('ES') && (pub_key = Company.where(id: payload['iss']).pluck(:public_key).first)
    # ECDSA token
    OpenSSL::PKey::EC.new(pub_key)

  else
    nil
  end
end.first
```

### DIFF
--- a/lib/jwt.rb
+++ b/lib/jwt.rb
@@ -141,6 +141,7 @@ module JWT
   def signature_algorithm_and_key(header, payload, key, &keyfinder)
     if keyfinder
       key = (keyfinder.arity == 2) ? keyfinder.call(header, payload) : keyfinder.call(header)
+      raise JWT::DecodeError, 'No verification key available'  unless key
     end
     [header['alg'], key]
   end
--- a/lib/jwt.rb
+++ b/lib/jwt.rb
@@ -122,7 +122,7 @@ module JWT
     merged_options = DEFAULT_OPTIONS.merge(custom_options)
     decoder = Decode.new jwt, key, verify, merged_options, &keyfinder
     header, payload, signature, signing_input = decoder.decode_segments
-    decode_verify_signature(key, header, signature, signing_input, merged_options, &keyfinder) if verify
+    decode_verify_signature(key, header, payload, signature, signing_input, merged_options, &keyfinder) if verify
     decoder.verify
 
     raise(JWT::DecodeError, 'Not enough or too many segments') unless header && payload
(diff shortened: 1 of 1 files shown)


CodeMonkeySteve:
Added check that `keyfinder` returns a key.

lib/jwt.rb
```diff
@@ -122,24 +122,27 @@ def decode(jwt, key = nil, verify = true, custom_options = {}, &keyfinder)
...
       raise JWT::IncorrectAlgorithm, 'Expected a different algorithm'
     end
     verify_signature(algo, key, signing_input, signature)
   end
 
-  def signature_algorithm_and_key(header, key, &keyfinder)
-    key = yield(header) if keyfinder
+  def signature_algorithm_and_key(header, payload, key, &keyfinder)
+    if keyfinder
+      key = (keyfinder.arity == 2) ? keyfinder.call(header, payload) : keyfinder.call(header)
```
excpt: Please change your code to use `yield` instead of keyfinder.call (`yield(header, playload)` and `yield(header)`) and use a full `if else` condition statement instead of your ternary conditions.

```ruby
key = if keyfinder.arity == 2
          yield(header, payload)
        else
          yield(header)
```

lib/jwt.rb
```diff
@@ -122,24 +122,27 @@ def decode(jwt, key = nil, verify = true, custom_options = {}, &keyfinder)
...
+      raise JWT::DecodeError, 'No verification key available'  unless key
```
excpt: Code smell: There is an additional space in front of the `unless` key word.

CodeMonkeySteve:
Made style changes.

lib/jwt.rb
CodeMonkeySteve: With the arity check, I prefer `call` to `yield`, as it makes it more obvious that `keyfinder` is the block being called.  But I'll defer to your style.

lib/jwt.rb
CodeMonkeySteve: Apparently no one likes my style of putting two spaces before postfix conditionals, but I wouldn't call it a code smell. ;)

excpt:
Thanks. :)

Issue #155, opened 2016-06-29
ruby-jwt::raw_to_asn1: Fails for signatures less than byte_size


joykoduru:
We use ruby-jwt in one of our applications and recently we had noticed that if the Signature passed to this method is shorter than byte_size, I returns an invalid signature which fails the verification and causing a 500 Internal Server Error. 

See the below sample code to understand better:

```
2.2.3 :004 > str = 'abcdef'
 => "abcdef"
2.2.3 :005 > str[0..10]
 => "abcdef"
2.2.3 :006 > str[10..-1]
 => nil
```

To fix this we could add a check for passed in signature’s length on [#L165](https://github.com/jwt/ruby-jwt/blob/v1.5.4/lib/jwt.rb#L165) and return “” (empty string) if signature is shorter.

If you feel this could be the right way to fix this issue, let us know how we can contribute to ruby-jwt. 

P.S: We are using the ES256 algorithm. It looks like our keys are 256 bits long.

Please let us know your thoughts.

excpt:
Please submit your PR. I will be happy to merge it.

PR #175, opened 2016-11-14
Return empty string if signature less than byte_size #155

xamenrax:
Is it what you wanted to accomplish in #155 ?


xamenrax:
Is it what you wanted to accomplish in #155 ?

excpt:
Thanks!

PR #183, opened 2017-01-04
Validate audience when payload is a scalar and options is an array

steti:
I recently updated from gem version 1.5.1 to 1.5.6 in one of my applications and I found that this behavior has changed. In my application, the JWT issuer creates the JWT with a scalar (string) `aud` claim.  The JWT consumer verifies the `aud` claim against an array of possible values. The consumer has multiple valid names and a valid JWT could target any of those names as an `aud`.  This works fine in v1.5.1, but the audience validation fails in v1.5.6.  I'm not sure if this change was intentional. Based on my reading of the RFC this particular case seems like an implementation detail.

### DIFF
--- a/lib/jwt/verify.rb
+++ b/lib/jwt/verify.rb
@@ -20,23 +20,11 @@ module JWT
     def verify_aud
       return unless (options_aud = extract_option(:aud))
 
-      if @payload['aud'].is_a?(Array)
-        verify_aud_array(@payload['aud'], options_aud)
-      else
+      if ([*@payload['aud']] & [*options_aud]).empty?
         raise(
           JWT::InvalidAudError,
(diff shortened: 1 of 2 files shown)


cgeers:
In addition to elegantly fixing the problem as described, the implementation here also addresses a separate problem I'm having validating the audience claim.

When both the JWT aud claim, and the aud options field are arrays, the current verification implementation in 1.5.6 requires that *all* aud options elements be present in the aud claim of the token, but this behavior doesn't make sense (at least in my use case). 

The changes proposed here appear to behave such that aud verification will pass when *any* of the aud options are present in the aud claim of the token.  This is a much more sensible behavior.

+1 for merger, LGTM

PR #187, opened 2017-02-05
Add options for claim-specific leeway

### LINKED ISSUE #129: The leeway parameter is applies to all time based verifications
Having expiration and not before timeouts using the same leeway is surprising and could lead JWTs to live longer than expected when both `exp` and `nbf` are being verified and a leeway is desired on one but not the other.

EmilioCristalli:
This PR intends to fix #129 by adding one independent leeway option for each claim.

I tried to make it backwards compatible by keeping the "global" `leeway` option (which is overwritten by each claim-specific leeway if they are present).

Let me know what you think

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ token = JWT.encode exp_payload, hmac_secret, 'HS256'
 
 begin
   # add leeway to ensure the token is still accepted
-  decoded_token = JWT.decode token, hmac_secret, true, { :leeway => leeway, :algorithm => 'HS256' }
+  decoded_token = JWT.decode token, hmac_secret, true, { :exp_leeway => leeway, :algorithm => 'HS256' }
 rescue JWT::ExpiredSignature
   # Handle expired token, e.g. logout user or deny access
 end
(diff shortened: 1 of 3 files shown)


excpt:
Hey @EmilioCristalli

Thanks for the PR.

Can you drop the leeway for the `iat` claim, please?

https://tools.ietf.org/html/rfc7519#section-4.1.6

The RFC does not allow leeways for that claim.

Only the `exp` and `nbf` claims are allowed to have a leeway.

EmilioCristalli:
Hey @excpt , I was also thinking about that, and my two concerns are:

1. **Backwards compatibility**: if we remove the leeway for `iat` we change a functionality people can be relying on, is it worth doing it for 2.0?
2. **`iat` validation**: I might be misinterpreting the RFC, but it seems like it doesn't talk about validating `iat`, it's like its purpose is to determine the age of the token, so maybe that's why it doesn't talk about a leeway for this claim.

In my opinion, the `iat` validation is an extra feature provided by the gem, so I would either remove completely this validation, or keep it with a leeway, as the rest of time sensitive validations have a leeway.

PR #195, opened 2017-02-12
Move signature logic to its own module

EmilioCristalli:
Not sure if this change was planned, but seemed aligned with the rest of modules the gem has.
Also I think this will help to start working on redefining the public API (I guess that is planned for version 2).

Let me know what you think, if you want me to make changes, or if it's just not the direction you guys wanted to go.

Thanks!

### DIFF
--- a/lib/jwt/signature.rb
+++ b/lib/jwt/signature.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+require 'openssl'
+
+module JWT
+  module Signature
+    extend self
+
+    NAMED_CURVES = {
+      'prime256v1' => 'ES256',
+      'secp384r1' => 'ES384',
+      'secp521r1' => 'ES512'
+    }.freeze
+
+    def sign(algorithm, msg, key)
+      if %w(HS256 HS384 HS512).include?(algorithm)
+        sign_hmac(algorithm, msg, key)
+      elsif %w(RS256 RS384 RS512).include?(algorithm)
+        sign_rsa(algorithm, msg, key)
+      elsif %w(ES256 ES384 ES512).include?(algorithm)
+        sign_ecdsa(algorithm, msg, key)
+      else
+        raise NotImplementedError, 'Unsupported signing method'
+      end
+    end
+
+    private
+
+    def sign_rsa(algorithm, msg, private_key)
+      raise EncodeError, "The given key is a #{private_key.class}. It has to be an OpenSSL::PKey::RSA instance." if private_key.class == String
+      private_key.sign(OpenSSL::Digest.new(algorithm.sub('RS', 'sha')), msg)
+    end
+
+    def sign_ecdsa(algorithm, msg, private_key)
+      key_algorithm = NAMED_CURVES[private_key.group.curve_name]
+      if algorithm != key_algorithm
+        raise IncorrectAlgorithm, "payload algorithm is #{algorithm} but #{key_algorithm} signing key was provided"
+      end
+
+      digest = OpenSSL::Digest.new(algorithm.sub('ES', 'sha'))
+      asn1_to_raw(private_key.dsa_sign_asn1(digest.digest(msg)), private_key)
+    end
+
+    def sign_hmac(algorithm, msg, key)
+      OpenSSL::HMAC.digest(OpenSSL::Digest.new(algorithm.sub('HS', 'sha')), key, msg)
+    end
+
+    def asn1_to_raw(signature, public_key)
+      byte_size = (public_key.group.degree + 7) / 8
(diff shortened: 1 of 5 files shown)


excpt:
Thank you very much!

This is exactly the way to go. 👍

lib/jwt/signature.rb
```diff
@@ -0,0 +1,105 @@
+# frozen_string_literal: true
+require 'openssl'
+
+module JWT
+  module Signature
+    extend self
```
excpt: Please use `module_function` instead of `extend self`

lib/jwt/signature.rb
```diff
@@ -0,0 +1,105 @@
...
+
+    HMAC_ALGORITHMS = %w(HS256 HS384 HS512).freeze
+    RSA_ALGORITHMS = %w(RS256 RS384 RS512).freeze
+    ECDSA_ALGORITHMS = %w(ES256 ES384 ES512).freeze
+
+    NAMED_CURVES = {
```
excpt: This hash has been moved into the JWT::DefaultOptions module. Please merge the current master into this PR.

lib/jwt/signature.rb
```diff
@@ -0,0 +1,105 @@
+# frozen_string_literal: true
+require 'openssl'
+
+module JWT
+  module Signature
+    extend self
+
+    HMAC_ALGORITHMS = %w(HS256 HS384 HS512).freeze
```
excpt: Please move these constants into the `DefaultOptions` module.

lib/jwt/signature.rb
EmilioCristalli: what about `private` functions? I don't think it's possible to define private function with `module_functions`. 
What are the benefits of `module_functions`?

EmilioCristalli:
Thanks for the comments @excpt 

This branch is already up-to-date with master, sorry I missed `NAMED_CURVES` was moved to the `DefaultOptions` module. 

In my opinion the constants `*_ALGORITHMS` and `NAMED_CURVES` make more sense in the `Signature` module. This module is the only one that uses them, and they don't seem to be configurable like the other options, so I would rather remove `NAMED_CURVES` from the `DefaultOptions` module.

Do you still think it's worth moving the constants to `DefaultOptions`?

excpt:
Agreed. Move the algorithms into the signature module. This makes way more sense. The rest is looking fine. After that change i'll merge your PR.

Thanks again. 🍻

lib/jwt/signature.rb
excpt: Using `module_functions` is the preferred coding style.

You're alway able to define a function as private by using this syntax:

```ruby
def my_fun(my_param) do
  puts my_param
end
private :my_fun
```

EmilioCristalli: I don't think `module_function` respects the private visibility.

```ruby
module A
  module_function

  def my_fun(my_param)
    puts my_param
  end
  private :my_fun
end

A.my_fun("calling private method")   #=>  "calling private method"
```

EmilioCristalli:
@excpt I just removed `NAMED_CURVES` in this commit 8ed32d7096948afd6cf2f77c7f5fc3cb0b7d1eac. 

Let me know what you want to do with the `module_function` vs `extend self` discussion

excpt:
After looking through the last refactoring to move logic out of the jwt.rb monolith I think the better solution to that problem is to rewrite the module as a class. The Encoding and Verify refactorings took the same approach. This will give you full control about what's private logic and what should be exposed to the outside.

EmilioCristalli:
I think normally a class is meant to have both behavior and state. In this case, `Signature` doesn't really need state, it's more like two util methods that would let you generate or verify a signature (given an algorithm and key).

So, if you implement `Signature` as a class you would either use instance methods and you would have to create a new instance to use the methods (which shouldn't be necessary because it doesn't have any state) or you use class methods (with private class methods too), which basically is the same as the `module` with `extend self`, but it feels like a `module` better defines the purpose of the util functions, because it doesn't hold state.

And I think `module` with `extend self` gives you as much control of private methods as a class (actually I think to define private class methods is a little trickier because you have to either use `class << self` or `private_class_method`)

@excpt do you still prefer to implement this as a class? would you use instance or class methods?

excpt:
I totally over-engineered the problem. Sorry.

Thank you very much for the contribution and your patience to withstand my pretty silly review. 👍 

The code solves the problem and this is what counts.

excpt:
@EmilioCristalli 
Please take a look at your PR #196 contribution. After merging this PR your new refactor PR has now merge conflicts.

EmilioCristalli:
no problem @excpt ! 

I just rebased and solved conflicts on #196

PR #203, opened 2017-04-26
Fix JWT.encode(nil)

tmm1:
After https://github.com/jwt/ruby-jwt/commit/9c44e72dd8e8c6d9809eeaf4b601d58a86886cf4 in #164, it is no longer possible to encode a nil value.

This PR fixes the issue for `nil`. But it occurs to me that similar problems could occur if someone tried to encode an array.


tmm1:
Any objections to this change? It's quite straightfoward...

Issue #206, opened 2017-05-29
Verifying expiration without leeway throws exception


fightingtheboss:
Both in `1.5.6` and in `master`, trying to verify the `exp` of a token without providing an explicit leeway throws the following exception:

`TypeError: nil can't be coerced into Integer`

This is because [this line](https://github.com/jwt/ruby-jwt/blob/master/lib/jwt/verify.rb#L34) (on master) relies on the return value from the options Hash to add time to the `exp` integer. If `leeway` isn't set in the options, the above error is thrown.

I think the leeway should default to 0 if not found, [as seems to be the assumption in the specs](https://github.com/jwt/ruby-jwt/blob/master/spec/jwt/verify_spec.rb#L8).

PR #223, opened 2017-09-01
Use a default leeway of 0

travisofthenorth:
This PR addresses https://github.com/jwt/ruby-jwt/issues/206. It sets up a hash of defaults (which simply has a `leeway` value). The spec I added fails with the error mentioned in the issue without this fix.

### DIFF
--- a/lib/jwt/verify.rb
+++ b/lib/jwt/verify.rb
@@ -4,6 +4,10 @@ require 'jwt/error'
 module JWT
   # JWT verify methods
   class Verify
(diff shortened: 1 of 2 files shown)


travisofthenorth:
@excpt I'd love to see this gem get past the 2.0.0 finish line, so please let me know if there is anything I can do to help.

excpt:
@travisofthenorth Thank you very much.

Working active on the 2.0.0 release. Estimated arrival 4th of September.

Issue #227, opened 2017-09-05
Why doesn't the decode function use a default algorithm?


madkin10:
Upgraded to 2.0.0 and started seeing this error everywhere I was calling `JWT.decode`
```
expected JWT::VerificationError, got #<JWT::IncorrectAlgorithm: An algorithm must be specified>:
```

`JWT.encode` defaults to the `HS256` algorithm, why not do the same for `JWT.decode`? It would be a simple change to the `DEFAULT_OPTIONS`.

```ruby
module JWT
  module DefaultOptions
    DEFAULT_OPTIONS = {
      verify_expiration: true,
      verify_not_before: true,
      verify_iss: false,
      verify_iat: false,
      verify_jti: false,
      verify_aud: false,
      verify_sub: false,
      leeway: 0,
      algorithm: 'HS256'
    }.freeze
  end
end

```

excpt:
The default algorithm was dropped with PR #184.

It is a security issue.

The documentation is wrong. 

#226 addresses this issue.

madkin10:
@excpt  The default algorithm was not dropped in #184. #184 changed the behavior to check the header's algorithm against the one supplied in options. Why not allow people to easily make the decision to use the default algorithm for decode? I don't think requiring someone to type in a string really improves security. The security improvement came from checking the header's algorithm against the one in options. Whether or not that is coming from a default value or a a passed in string it still applies the same level of security.

fabn:
Got the same issue, if you wan't to provide that behaviour without changing all your codebase you can use this monkeypatch:

```ruby
module JwtAlgorithmChooser
  # Default algorithm for decoding
  DEFAULT_ALGORITHM = 'HS256'
  # Automatically choose decoding algorithm when not given
  # @see https://github.com/jwt/ruby-jwt/issues/227
  def decode(jwt, key = nil, verify = true, custom_options = {}, &keyfinder)
    super(jwt, key, verify, custom_options.merge(algorithm: DEFAULT_ALGORITHM), &keyfinder)
  end
end
# Monkey patch the class until #227 is solved
# @see https://stackoverflow.com/a/32334444/518204
JWT.singleton_class.prepend(JwtAlgorithmChooser)
```

excpt:
First of all: Sorry for the delay. I'm back from vacation.

@madkin10 I merged your PR.

I think I will get the 2.1.0 release ready until Friday.

Thanks for all the feedback and contribution.

madkin10:
[...]

PR #228, opened 2017-09-05
Add HS256 algorithm to decode default options

madkin10:
# Things Done
- Added `algorithm: 'HS256'` to the JWT.decode default options list
- Added spec tests

Addresses Issue: #227


excpt:
As mentioned in #227. This is a documentation issue. This beahvior is intended as of #184.

Issue #217, opened 2017-08-18
Ed25519 support planned?


Justin-Maxwell:
Hi.

Are there plans to incorporate Ed25519?  (via RbNaCl presumably).

Cheers.

excpt:
@Justin-Maxwell 

Sure, this would be an awesome addition. For a more complete JWT spec implementation.

Setting this to "up for grabs"

excpt:
@Justin-Maxwell This feature will be added with 2.1.0. See #229.

PR #229, opened 2017-09-06
Support for ED25519

ab320012:
Added support + tests, readme section for eddsa algo ed25519 
https://github.com/jwt/ruby-jwt/issues/217

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,8 +149,7 @@ puts decoded_token
 * ED25519 
 
 ```ruby 
-private_key = RbNaCl::Signatures::Ed25519::SigningKey.new("abcdefghijklmnopqrstuvwxyzABCDEF") # the
-new method takes a 32 bit string
+private_key = RbNaCl::Signatures::Ed25519::SigningKey.new("abcdefghijklmnopqrstuvwxyzABCDEF")
 public_key = private_key.verify_key
 token = JWT.encode payload, private_key, 'ED25519' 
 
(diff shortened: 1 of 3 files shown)


ab320012:
method length + complexity for verify and arity of sign + verify eddsa methods follow the surrounding code convention. I can probably fix these but would have to modify existing code, let me know if this is ok

excpt:
@ab320012 Thank you for the PR!

Refactoring the code to reduce the arity and other code smells should happen in a different PR. Don't do it in this PR.

Reviewing you code now.

ab320012:
@excpt Thanks, will add that one after this one closes. Wondering if class based solution will make more sense. 

```ruby 
# lib/jwt/algos/hmac.rb
class Hmac
   # other classes same structure: ecdsa, eddsa, rsa
  TYPES= %w(HS256 HS512256 HS384 HS512).freeze 
 def initialize; end 
 def sign (signing_input); end
 def verify(verifying_input); end 
end

# signing_input and verifying_input will be refactored into a struct 

#lib/signature.rb 
Signed = Struct.new(:algo, :key, :msg) 
Verified = Struct.new(:algorithm, :public_key, :signing_input, :signature)
```

Let me know if something like this is ok with you for next PR

excpt:
@ab320012 That's the way to go. Most important: the two signatures of the main methods `JWT.encode` and `JWT.decode` need to stay the same. Everything else is code behind and that should not break anything for the users.

PR #231, opened 2017-09-08
Show backtrace if rbnacl-libsodium not loaded

buzztaiki:
I think that backtrace should be shown to make to solve problem as easy when rbnacl-libsodium is not loaded.


excpt:
@buzztaiki Thanks for the PR. This really makes the usage a lot easier.

PR #234, opened 2017-09-15
Use case-compare for more flexibility in sub verification

bmalinconico:
I had a need to verify only the prefix of the sub claim. This PR changes the validation to perform a case compare, instead of casting to a string and doing exact equality.


excpt:
@bmalinconico I close this PR because #241 implemented the scenario better. Thanks for your contribution.

PR #236, opened 2017-09-18
Fix string payload issue

excpt:


excpt:
@ShockwaveNN, @ab320012 If you can spare a view moments to review this changed, please.

Issue #207, opened 2017-06-22
Verify JTI Proc


MPT-ladams:
I'm attempting to use this logic to set & verify the JTI value:

```ruby
jti_raw = [hmac_secret, iat].join(':').to_s
jti = Digest::MD5.hexdigest(jti_raw)
```

When my verify_jti proc is called, I only have access to the @payload['jti'] value passed. How can I access the issued at value from the payload to verify?

Or is this not the intention of how the provided proc is to be used?

Thanks!

ab320012:
@excpt will the following work? (adding iat var and passing to proc) 
```ruby
# jwt/verify.rb
    def verify_jti
      options_verify_jti = @options[:verify_jti]
      jti = @payload['jti']
      iat = @payload['iat'] 
      if options_verify_jti.respond_to?(:call)
        raise(JWT::InvalidJtiError, 'Invalid jti') unless options_verify_jti.call(jti, iat)
      elsif jti.to_s.strip.empty?
        raise(JWT::InvalidJtiError, 'Missing jti')
      end
    end

```

excpt:
@MPT-ladams Would that be a fitting solution to access the payload of the token?

PR #238, opened 2017-09-19
verify takes 2 params, second being payload closes: #207

ab320012:
Added optional payload to verify_jti + passing tests for payload and arguements #207


excpt:
@ab320012 Thanks for the PR!

ab320012:
@excpt  Thanks Tim!

PR #239, opened 2017-09-26
Support dynamic algorithm selection

CodeMonkeySteve:
This is a follow-up to #172, which was intended to allow `keyfinder` to determine the decode key dynamically.  Adding the requirement to specify the algorithm causes a regression here, as the algorithm isn't known ahead of time.
There are a few ways to address this, but I think the simplest is just to allow the `algorithm` option to contain an array of supported algorithms, and only error if the request algorithm isn't in the list.


excpt:
@CodeMonkeySteve thanks for the contribution. Can you add a test for that scenario, please?

excpt:
@CodeMonkeySteve I close this PR. #241 implemented the whole scenario in a better way, IMHO.

Thanks for your contribution.

PR #248, opened 2017-11-18
Fix link format

y-yagi:
Because there is unnecessary space, markdown link format is not generated.
Sorry if it was intentional.


excpt:
@y-yagi Thanks for the PR. Merged.

y-yagi:
Thanks!

PR #250, opened 2017-12-27
Fix README

rono23:
I think `sub` should be a symbol, not string.


excpt:
@rono23  Thanks for the PR.
@Spone Thanks for the review.

I know the documentation is not up to date.

I recommend using the preferred hash syntax.

```ruby
{key: value}
```

Spone:
Yup, even better :ok_hand:

rono23:
@Spone @excpt Thanks you for the review! I updated my commit. Please review again :bow:

PR #257, opened 2018-03-22
Removed leeway from verify_iat

### LINKED ISSUE #247: Issued at validation
Hey guys, first off: thanks for this great library!

I have a common question, or maybe I don't get the point right now. So lets say a token was 
issued a hour before now. (Say 1pm) And now is 2pm. Why is the verify_iat checking if iat is
newer than now? 

```ruby
if .. @payload['iat'].to_f > (Time.now.to_f + leeway)
  raise ..
end
```

I mean yeah, the leeway could be a hour or more, but isn't this exactly what nbf (not before) 
should take care of? I found this "issue" (if I got it right) on a common test case. (iat -1h, exp: +2h, now is in between)

Could someone explain this to me, or confirm a bug?

ab320012:
Fixes #247.

### DIFF
--- a/spec/jwt/verify_spec.rb
+++ b/spec/jwt/verify_spec.rb
@@ -87,6 +87,11 @@ module JWT
         Verify.verify_iat(payload, options)
       end
 
+      it 'must ignore configured leeway' do
+        expect{Verify.verify_iat(payload.merge('iat' => (iat + 60)), options.merge(leeway: 70)) }
+          .to raise_error(JWT::InvalidIatError)
+      end
+
       it 'must properly handle integer times' do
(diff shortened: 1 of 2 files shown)


spec/jwt/verify_spec.rb
```diff
@@ -87,14 +87,6 @@ module JWT
         Verify.verify_iat(payload, options)
       end
 
-      it 'must allow configured leeway' do
```
excpt: It would be better to invert one of these tests to ensure that the leeway is ignored.

excpt:
@ab320012 Thank you for the PR! I added my review. :) Let me know if you have the time to improve that or if any help is needed.

Jack12816:
Thanks, guys!

PR #263, opened 2018-04-19
fixed ECDSA supported versions const

starbeast:


excpt:
Hi @starbeast,

Thanks for spotting and fixing this code smell. 👍

PR #266, opened 2018-05-04
improving code quality of jwt module

ab320012:
I started looking into the #errors and #value object return for decode and wanted to clean up the logic before implementing

### DIFF
--- a/lib/jwt.rb
+++ b/lib/jwt.rb
@@ -22,42 +22,59 @@ module JWT
     encoder.segments
   end
 
-  def decode(jwt, key = nil, verify = true, custom_options = {}, &keyfinder)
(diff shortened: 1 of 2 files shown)


ab320012:
@excpt I am thinking about implementing DecodedToken class that will return from JWT#decode. It will act as array with two elements for backward compatability. It should also probably implement a payload and header class to access the data directly, as  well as an errors method. My issue is not sure how to implement the backwards compatibility for errors. if we return an array from #errors method on DecodedToken instead of throwing errors it will break all existing implementations, let me know.

excpt:
@ab320012 The introduction of Encode/Decode value objects should happen in a different PR.

Thanks for all your effort and time you invest!

Issue #282, opened 2018-09-07
Inconsistent handling of payload claim data types


jamesstonehill:
When passing in a `Time` object as the value of the `exp` claim, if the key of the payload is a string it raises a `JWT::InvalidPayload` error. When using the symbol as the key, it does not raise the error. 

### When the `exp` claim key is a string
```ruby
exp_payload = { 'exp' => Time.now }
token = JWT.encode exp_payload, 'asdf'
# => JWT::InvalidPayload: exp claim must be an integer
```

### When the `exp` claim key is a symbol
```ruby
exp_payload = { exp: Time.now }
token = JWT.encode exp_payload, 'asdf'
# => "eyJhbGciOiJIUzI1NiJ9.eyJleHAiOiIyMDE4LTA5LTA3IDEzOjA3OjU0ICswMTAwIn0.9PxmdZFISuVXQQTDA6bw4vavK428o0R1WvcIZ5np6Lg"
```

Symbol and string keys in the payload should be handle in a consistent way.

excpt:
Hi @jamesstonehill,

thanks for the detailed report. Can you specify your ruby-jwt version you're using?

jamesstonehill:
@excpt yeah, sorry I should have included that in the report. I'm using the latest release, 2.1.0.

excpt:
@jamesstonehill can you check against the current master branch if #293 fixed this issue for you, please?

jamesstonehill:
@excpt yeah this is fixed in the current master.

PR #287, opened 2018-09-18
Claims Validation

jamesstonehill:
In an effort to make the claim validation a little more extendable and to fix [this issue](https://github.com/jwt/ruby-jwt/issues/282), I've added a `ClaimsValidator` class that is responsible for validating validatable claims.

I thought that it is a little inconsistent that we only validate the `exp` claim is an integer even though the JWT RFC mandates that all time (iat, nbf, and exp) claims "MUST be a number containing a NumericDate value".

### DIFF
--- a/lib/jwt/claims_validator.rb
+++ b/lib/jwt/claims_validator.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require_relative './error'
+
+module JWT
+  class ClaimsValidator
+    def initialize(payload)
+      @payload = payload.each_with_object({}) { |(k, v), h| h[k.to_sym] = v }
+    end
+
+    def validate
+      validate_exp if @payload[:exp]
+
+      true
+    end
+
+    private
+
+    def validate_exp
+      raise InvalidPayload, 'exp claim must be an integer' unless @payload[:exp].is_a?(Integer)
+    end
+  end
+end
--- a/lib/jwt/encode.rb
+++ b/lib/jwt/encode.rb
@@ -1,52 +1,76 @@
 # frozen_string_literal: true
 
-require 'json'
-require_relative './claims_validator'
-
 # JWT::Encode module
 module JWT
   # Encoding logic for JWT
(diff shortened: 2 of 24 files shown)


lib/jwt/claims_validator.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require_relative './error'
+
+module JWT
+  class ClaimsValidator
+    def initialize(payload)
+      @payload = payload.each_with_object({}) { |(k, v), h| h[k.to_sym] = v }
```
jamesstonehill: Unfortunately `simbolize_keys` is too recent to use 😢

lib/jwt/encode.rb
```diff
@@ -9,7 +10,7 @@ class Encode
     attr_reader :payload, :key, :algorithm, :header_fields, :segments
 
     def self.base64url_encode(str)
-      Base64.encode64(str).tr('+/', '-_').gsub(/[\n=]/, '')
+      Base64.urlsafe_encode64(str).delete('=')
```
jamesstonehill: This does the same thing as `Base64.urlsafe_encode64(str, padding: false)` but the padding option wasn't added until recently.

jamesstonehill:
@excpt seems like the specs are failing for 2.2 because of a dependency mismatch with openssl

lib/jwt/claims_validator.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    def validate
+      validate_exp if @payload[:exp]
```
anakinj: You have to check that the key actually is present in the payload, this row is now checking if the value returned from the hash is thruthy. 

I think this will not validate cases where exp is false or nil eg:
```
{ exp: false }
```

jamesstonehill:
Switching branches to this one for a cleaner commit history. https://github.com/jwt/ruby-jwt/compare/master...jamesstonehill:claims-validation?expand=1

PR #294, opened 2019-01-21
Fix JSON conflict between #293 and #292

anakinj:
Hi, there were some conflicting changes between #293 and #292. This PR should fix that.


anakinj:
@excpt, there were some incompatibilities between the lates two branches merged to master. This change should make those compatible with each other.

PR #297, opened 2019-02-11
Make sure alg parameter value isn't added twice

korstiaan:
The `alg` parameter key (`ALG_KEY`) is a string, and added to the header hash as such (`{ "alg" =>  "foo" }`). If it's already in the header hash, but as a symbol (`{ alg: "foo"}), it's appended to the this hash (`{ alg: "foo", "alg" => "foo"}`) resulting in the end in a different token.

We noticed Apple's Mapkit JS complaing about this.

This PR fixes this.

### DIFF
--- a/lib/jwt/encode.rb
+++ b/lib/jwt/encode.rb
@@ -39,7 +39,7 @@ module JWT
     end
 
     def encode_header
-      encode(@headers.merge(ALG_KEY => @algorithm))
+      encode(@headers.transform_keys(&:to_s).merge(ALG_KEY => @algorithm))
     end
 
     def encode_payload
--- a/lib/jwt/encode.rb
+++ b/lib/jwt/encode.rb
@@ -39,7 +39,13 @@ module JWT
     end
 
     def encode_header
-      encode(@headers.transform_keys(&:to_s).merge(ALG_KEY => @algorithm))
+      stringified_header = {}
+
+      @headers.each do |key, value|
+        stringified_header[key.to_s] = value
+      end
+
+      encode(stringified_header.merge(ALG_KEY => @algorithm))
     end
 
     def encode_payload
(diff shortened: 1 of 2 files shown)


lib/jwt/encode.rb
```diff
@@ -39,7 +39,7 @@ def encoded_header_and_payload
     end
 
     def encode_header
-      encode(@headers.merge(ALG_KEY => @algorithm))
+      encode(@headers.transform_keys{ |key| key.to_s }.merge(ALG_KEY => @algorithm))
```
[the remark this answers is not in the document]
korstiaan: Fixed!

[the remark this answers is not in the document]
korstiaan: Automatically fixed by the fix above.

lib/jwt/encode.rb
```diff
@@ -39,7 +39,7 @@ def encoded_header_and_payload
     end
 
     def encode_header
-      encode(@headers.merge(ALG_KEY => @algorithm))
+      encode(@headers.transform_keys(&:to_s).merge(ALG_KEY => @algorithm))
```
anakinj: I think Hash#transform_keys was added to Ruby 2.5.0, so this won't work on older rubies. I think something like what is done [here](https://github.com/jwt/ruby-jwt/blob/master/lib/jwt/claims_validator.rb#L12) could be applied for the headers also.

Also if a copy of the ```@headers``` hash is created the ```Hash#merge!``` method could be used to save one object allocation.

korstiaan: I've moved the stringify'ing (as done in the claims validator) to the constructor, and directly add `ALG_KEY` to the headers hash instead of `.merge`

anakinj: Nice, looks good to me.

PR #305, opened 2019-04-08
Make JWT::Signature.verify return true on success

bdewater:
Before this change it returns nil, this makes it easier to write code like
```ruby
if JWT::Signature.verify(...)
  do_the_thing
end
```

### DIFF
--- a/lib/jwt/signature.rb
+++ b/lib/jwt/signature.rb
@@ -43,6 +43,7 @@ module JWT
       end
       verified = algo.verify(ToVerify.new(algorithm, key, signing_input, signature))
       raise(JWT::VerificationError, 'Signature verification raised') unless verified
+      true
     rescue OpenSSL::PKey::PKeyError
       raise JWT::VerificationError, 'Signature verification raised'
     ensure


bdewater:
@excpt I'd appreciate any feedback you'd have 🙂

anakinj:
Im wondering if this method will ever return anything else than `true` and in all the other cases raise an exception?

You example could be just without the nesting.
```
JWT::Signature.verify(...)
do_the_thing
```

It's maybe not clear (and I'm not sure) that the `verify` method will always raise an exception when verification fails. Maybe we should make an `verify!` an alias to that method that would indicate that it always rises when the token is not valid.

excpt:
> Im wondering if this method will ever return anything else than `true` and in all the other cases raise an exception?
> 
> You example could be just without the nesting.
> 
> ```
> JWT::Signature.verify(...)
> do_the_thing
> ```
> 
> It's maybe not clear (and I'm not sure) that the `verify` method will always raise an exception when verification fails. Maybe we should make `verify!` an alias to that method that would indicate that it always rises when the token is not valid.

You are correct. 

This method will always raise an exception on error conditions.

This would be an improvement for a future release since it'll break the current behaviour.

I'll schedule this one for a 3.0.0 release.

anakinj:
Im going to close this, gotten pretty stale already. Lets try to come up with a great api for the version 3.x

Issue #59, opened 2015-02-25
Support verifying signature signed using x5c header


punkle:
It is currently possible to verify the signature of a jwt token using the x5c header by providing a key finder function. e.g.

```ruby
jwt_token = ::JWT.decode(token) do |headers|
  OpenSSL::X509::Certificate.new((headers['x5c'].first)).public_key
end
```

In order to validate the x5c certificate chain in full, I must provide my own implementation. It would be great if ruby-jwt gem could support the validation of the x5c certificate chain natively.

Here is an excerpt from https://tools.ietf.org/html/draft-ietf-jose-json-web-key-41#section-4.7

> 4.7.  "x5c" (X.509 Certificate Chain) Parameter
> 
>   The "x5c" (X.509 Certificate Chain) member contains a chain of one or
>   more PKIX certificates [RFC5280].  The certificate chain is
>   represented as a JSON array of certificate value strings.  Each
>   string in the array is a base64 encoded ([RFC4648] Section 4 -- not
>   base64url encoded) DER [ITU.X690.1994] PKIX certificate value.  The
>   PKIX certificate containing the key value MUST be the first
>   certificate.  This MAY be followed by additional certificates, with
>   each subsequent certificate being the one used to certify the
>   previous one.  The key in the first certificate MUST match the public
>   key represented by other members of the JWK.  Use of this member is
>   OPTIONAL.

excpt:
We need to implement JWK for that. You may have a look at the 2.x branch.

stunney:
Don't forget, you need the ENTIRE certificate chain in that field.  The signing certificate *first* followed by each intermediate and root CA that owns that one.

punkle:
@stunney It 'MAY' contain the entire certificate chain. The last certificate in the x5c chain could be rooted in a  certificate store managed by the entity validating the token.

stunney:
You are correct.  The first must be the cert that signed the token.  The rest are optional.

aj-michael:
adding to this, using a keyfinder block is really ugly if you have additionaly options to pass in.

```ruby
JWT.decode(raw_token, nil, true, additional_options) do |header|
  ...
end
```

I don't like that I have to pass in a key (nil) that is going to be ignored anyways. Especially since I have to pass in an options has to do full validation in accordance with the RFC.

itstehkman:
Hey @punkle, I'm taking a quick stab at this since I need support for this. Open question:

[...]

PR #308, opened 2019-05-04
Support x5c JWT header as specified in RFC-7515 (JWS)

itstehkman:
Taking a stab at supporting x5c (https://github.com/jwt/ruby-jwt/issues/59)

https://tools.ietf.org/html/rfc7515#section-4.1.6

> The "x5c" (X.509 certificate chain) Header Parameter contains the
> X.509 public key certificate or certificate chain [RFC5280]
> corresponding to the key used to digitally sign the JWS.  The
> certificate or certificate chain is represented as a JSON array of
> certificate value strings.  Each string in the array is a
> base64-encoded (Section 4 of [RFC4648] -- not base64url-encoded) DER
> [ITU.X690.2008] PKIX certificate value.  The certificate containing
> the public key corresponding to the key used to digitally sign the
> JWS MUST be the first certificate.  This MAY be followed by
> additional certificates, with each subsequent certificate being the
> one used to certify the previous one.  The recipient MUST validate
> the certificate chain according to RFC 5280 [RFC5280] and consider
> the certificate or certificate chain to be invalid if any validation
> failure occurs.  Use of this Header Parameter is OPTIONAL.

I'll fix the ebert issues later, I just want the idea validated so I can move forward.


lib/jwt/x509/validator.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      def valid?
+        return true if @cert_chain.blank?
+
+        store = OpenSSL::X509::Store.new
+        @cert_chain.each do |cert|
+          store.add_cert cert
+        end
+        store.verify @signing_cert
+      end
```
bdewater: This block of code has security problems. From [RFC 5280 section 3.2](https://tools.ietf.org/html/rfc5280#section-3.2):

> In general, a chain of multiple certificates may be needed, comprising a certificate of the public key owner (the end entity) signed by one CA, and zero or more additional certificates of CAs signed by other CAs.  Such chains, called certification paths, are required because a public key user is only initialized with a limited number of assured CA public keys.

More concretely:
- Chain validation can not be skipped: this would allow an attacker to just strip the chain and inject their own certificate and use it to sign, bypassing any security check. Even a single certificate must chain correctly to a trust anchor.
- The trust anchor should be configurable by the consuming app. It may be tempting to default to `OpenSSL::X509::Store#set_defaults_path` but instead of allowing all the certificate authorities to sign, the user should specify this themselves and preferably only limit to root certificate(s) that apply for their use case. E.g. when validating a JWT from Google, use only the cert(s) from https://pki.goog/
- [RFC 5280 section 3.3](https://tools.ietf.org/html/rfc5280#section-3.3) also mentions revocation checking, which is absent in this PR.

I've been implementing all of the above in the WebAuthn gem, you can find my code here: https://github.com/cedarcode/webauthn-ruby/pull/208 - feel free to use it, I'd appreciate a [mention as co-author](https://help.github.com/en/articles/creating-a-commit-with-multiple-authors) if you do 🙂

itstehkman: @bdewater sorry for the late response here! Thank you for all the feedback here, definitely seems like I was missing some parts. And that webauthn PR looks impressive.

Unfortunately I am no longer working on this PR, as my project's requirements changed and I don't need to use `x5c` for JWTs anymore. Please feel free to take my work and continue it if you'd like!

rahulbajaj0509: @bdewater would you be interested in taking over this PR, because it looks like a important feature. Newer versions of the IDP's are supporting the x5c  header and I think we should support it in this gem too. 

Let me know if you would be able to work on this one, else I can pick it up and resolve :)

bdewater: I'm currently on vacation but was planning to pick this up again after I'm back next week :)

bdewater: See https://github.com/jwt/ruby-jwt/pull/338

PR #312, opened 2019-05-17
Improvement/encode hmac without key

JotaSe:
# Description

  For `hmac` strategy, will raise an exception if the secret key is `nil` but no when secret key is blank, for both cases should be the same.

# Analysis

  in `lib/jwt/algos/hmac.rb`the key value is used for 
```ruby
SecurityUtils.rbnacl_fixup(algorithm, key.to_s)
```
and 
```ruby
OpenSSL::HMAC.digest(OpenSSL::Digest.new(algorithm.sub('HS', 'sha')), key.to_s, msg)
```

in both cases, the key should be a `string` for https://ruby-doc.org/stdlib-2.4.0/libdoc/openssl/rdoc/OpenSSL/HMAC.html and
```ruby
# lib/jwt/security_utils.rb:49
# lib/jwt/security_utils.rb:53
key.bytesize
``` 

 # Solution:

in `JWT::Algos::Hmac#sign` set `key ||= ''` to convert `nil` to `String`

# QA

* Run tests
* `bin/console.rb` =>
```ruby
# Without secret key
token = JWT.encode payload, nil, 'HS256'

# eyJhbGciOiJIUzI1NiJ9.eyJkYXRhIjoidGVzdCJ9.pVzcY2dX8JNM3LzIYeP2B1e1Wcpt1K3TWVvIYSF4x-o
puts token

decoded_token = JWT.decode token, nil, true, { algorithm: 'HS256' }

# Array
# [
#   {"data"=>"test"}, # payload
#   {"alg"=>"HS256"} # header
# ]
puts decoded_token
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,21 @@ decoded_token = JWT.decode token, hmac_secret, true, { algorithm: 'HS256' }
 #   {"alg"=>"HS256"} # header
 # ]
 puts decoded_token
+
+# Without secret key
+token = JWT.encode payload, nil, 'HS256'
+
+# eyJhbGciOiJIUzI1NiJ9.eyJkYXRhIjoidGVzdCJ9.pVzcY2dX8JNM3LzIYeP2B1e1Wcpt1K3TWVvIYSF4x-o
+puts token
+
+decoded_token = JWT.decode token, nil, true, { algorithm: 'HS256' }
+
+# Array
+# [
+#   {"data"=>"test"}, # payload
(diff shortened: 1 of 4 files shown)


rahulbajaj0509:
@JotaSe good catch! mind squashing the commits to one?

JotaSe:
Sure @rahulbajaj0509

JotaSe:
Done @rahulbajaj0509

Issue #313, opened 2019-05-21
Importing JWK then exporting results in different `kid`


xcskier56:
If you import a JWK from json and then export it back to JSON, it is exported with a changed kid. This means that the JWKs loader is failing to find the key by its id.

Gem Version: 2.2.0.pre.beta.0
Ruby Version: 2.6.3

    key_hash = {
      kty: "RSA",
      kid: "xQbOc1Ydqq3vemT9WN4GWuOZKOFSPXq1xzXxON_yWFg",
      e: "AQAB",
      n: "unkoQzE-VJwehDczS5o5ijyaAH2JyzhdLd7oSndnS63gzFRXS_LpDq62JLGHMFCYo11VYRb9kn2cdlfG0IANqTt9SWJ-iSKmdRnZIm-q5XCJGmW7C1Gazh9kgmMeWq3Zq5cv83r0e-KkcmWiT_XMrXutTTtUAVms1bCoGCEq7NBb5mnGeFxK51yxlNTnSMk0SJBuYvJ_13XISUQE-hwCdtd9LyDVFcggJcDejWlO3EIAtsWS4aPBt-ln9y9wIVpmsmjmsRTmXtS8B3KH6f6qAPuxzTmCNWLp_jh-W60o3lmgkMaqwdc0OjUpc6Lhn9E9LzqftXKpV3H6QRQgicQmyQ",
      use: "sig",
      alg: "RS256"
    }

    jwk = JWT::JWK.import(key_hash)

    export = jwk.export

    key_hash[:kid] == export[:kid] # FALSE

    key_hash[:kid] # => "xQbOc1Ydqq3vemT9WN4GWuOZKOFSPXq1xzXxON_yWFg"
    export[:kid] # => "68945d6bbec4cf6e228c3675671dbd39b5686425bb075285dd2d19593c48000a"

anakinj:
This is fixed in master: https://github.com/jwt/ruby-jwt/commit/83b6e61fc4a2b80f4952b992ebc5f63a6b906441

PR #320, opened 2019-06-10
Preserve JWK kid on import

rkmetzl:
From issue #313 

The JWK `import` method generates a unique kid on export, but doesn't take the imported kid into account. The logic here allows an imported kid to be included on export, while allowing a kid to be generated if none is present.

### DIFF
--- a/lib/jwt/jwk/rsa.rb
+++ b/lib/jwt/jwk/rsa.rb
@@ -8,19 +8,21 @@ module JWT
       extend Forwardable
 
       attr_reader :keypair
+      attr_reader :jwk_kid
 
       def_delegators :keypair, :private?, :public_key
 
       BINARY = 2
       KTY    = 'RSA'.freeze
 
-      def initialize(keypair)
+      def initialize(keypair, kid = nil)
         raise ArgumentError, 'keypair must be of type OpenSSL::PKey::RSA' unless keypair.is_a?(OpenSSL::PKey::RSA)
-
+        @jwk_kid = kid
         @keypair = keypair
       end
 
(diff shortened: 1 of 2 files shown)


lib/jwt/jwk/rsa.rb
```diff
@@ -8,19 +8,21 @@ class RSA
...
-      def initialize(keypair)
+      def initialize(keypair, kid = nil)
         raise ArgumentError, 'keypair must be of type OpenSSL::PKey::RSA' unless keypair.is_a?(OpenSSL::PKey::RSA)
-
+        @jwk_kid = kid
         @keypair = keypair
       end
 
       def kid
+        return jwk_kid if jwk_kid
```
anakinj: Would it make sense to have this method as something like:
```
def kid
  @kid ||= generate_from_public_key
end
```

Then we can just set the instance variable ```@kid``` in the constructor.

volmer:
Hey @rkmetzl are you still planning on moving forward with things change? It would be great if we could set our own `kid` when instantiating JWKs 🙏

anakinj:
@excpt I think we could merge this as it is and then improve. There are some exciting things going on with the JWK support at the moment and this change would help us forward.

anakinj:
@excpt I think we could merge this as it is now and then make some improvements. There are some exciting changes related to JWKs happening.

phlegx:
This can be closed because it get solved with PR #375.

anakinj:
The initial fix by you @rkmetzl was included in #375. Thanks for your contribution.

Issue #332, opened 2019-09-13
JWK does not decode.


rahulbajaj0509:
while using the ruby version 2.3.5, I get this error:
```ruby
2.3.5 :010 > JWT.decode(token, nil, true, { algorithms: ['RS512'], jwks: jwk_loader})
NoMethodError: undefined method `set_key' for #<OpenSSL::PKey::RSA:0x0000000001e2c9e8>
	from /home/vagrant/.rvm/gems/ruby-2.3.5/gems/jwt-2.2.1/lib/jwt/jwk/rsa.rb:40:in `import'
	from /home/vagrant/.rvm/gems/ruby-2.3.5/gems/jwt-2.2.1/lib/jwt/jwk.rb:19:in `import'
	from /home/vagrant/.rvm/gems/ruby-2.3.5/gems/jwt-2.2.1/lib/jwt/jwk/key_finder.rb:19:in `key_for'
	from /home/vagrant/.rvm/gems/ruby-2.3.5/gems/jwt-2.2.1/lib/jwt/decode.rb:37:in `verify_signature'
	from /home/vagrant/.rvm/gems/ruby-2.3.5/gems/jwt-2.2.1/lib/jwt/decode.rb:26:in `decode_segments'
	from /home/vagrant/.rvm/gems/ruby-2.3.5/gems/jwt-2.2.1/lib/jwt.rb:28:in `decode'
	from (irb):10
	from /home/vagrant/.rvm/rubies/ruby-2.3.5/bin/irb:11:in `<main>'
```

Can anyone please help me out with this? or can suggest me with a workaround?

rahulbajaj0509:
@excpt any suggestions?

rahulbajaj0509:
Can we add an implementation our gem similar to this one? https://github.com/tootsuite/ostatus2/pull/3/files

rahulbajaj0509:
@excpt can we do something like this:
```
imported_key = OpenSSL::PKey::RSA.new  
if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.4.0') 
  def set_key(key, modulus, exponent)
    key.n = OpenSSL::BN.new(::Base64.urlsafe_decode64(jwk_data[:n]), BINARY)
    key.e = OpenSSL::BN.new(::Base64.urlsafe_decode64(jwk_data[:e]), BINARY)
  end
else
  def set_key(key, modulus, exponent)
[...]

PR #333, opened 2019-09-17
Support RSA.import for all Ruby versions.

rahulbajaj0509:
For Ruby versions < 2.4.0:
```
2.3.5 :006 > OpenSSL::PKey::RSA.new.respond_to?(:set_key)
 => false
```

This patch makes sure that RSA.import method is supported for all versions of Ruby.


excpt:
Fixes #332.

PR #338, opened 2019-11-04
Add x5c header key finder

bdewater:
This validates the certificate chain in accordance with RFC 5280, as described in [RFC 7515 section 4.1.6](https://tools.ietf.org/html/rfc7515#section-4.1.6).

To use this in your app, here are a couple of notes:
- you will want to cache the relevant CRL file(s) and make sure the cache is expired by the `OpenSSL::X509::CRL#next_update` timestamp
- in case you need to dynamically extract the CRL distribution point URIs from the x5c certificates, it is possible to use the `X5cKeyFinder` class directly in the keyfinder block argument passed to `JWT.decode`

Closes https://github.com/jwt/ruby-jwt/issues/59
Closes https://github.com/jwt/ruby-jwt/pull/308

### DIFF
--- a/lib/jwt/decode.rb
+++ b/lib/jwt/decode.rb
@@ -24,6 +24,7 @@ module JWT
       validate_segment_count!
       if @verify
         decode_crypto
+        verify_algo
         set_key
         verify_signature
         verify_claims
@@ -46,11 +47,13 @@ module JWT
       raise(JWT::VerificationError, 'Signature verification failed')
     end
 
-    def set_key
+    def verify_algo
       raise(JWT::IncorrectAlgorithm, 'An algorithm must be specified') if allowed_algorithms.empty?
       raise(JWT::IncorrectAlgorithm, 'Token is missing alg header') unless algorithm
       raise(JWT::IncorrectAlgorithm, 'Expected a different algorithm') unless options_includes_algo_in_header?
+    end
 
+    def set_key
       @key = find_key(&@keyfinder) if @keyfinder
       @key = ::JWT::JWK::KeyFinder.new(jwks: @options[:jwks]).key_for(header['kid']) if @options[:jwks]
       if (x5c_options = @options[:x5c])
--- a/lib/jwt/decode.rb
+++ b/lib/jwt/decode.rb
@@ -4,6 +4,7 @@ require 'json'
 
 require 'jwt/signature'
 require 'jwt/verify'
+require 'jwt/x5c_key_finder'
 # JWT::Decode module
 module JWT
   # Decoding logic for JWT
@@ -52,6 +53,9 @@ module JWT
 
       @key = find_key(&@keyfinder) if @keyfinder
       @key = ::JWT::JWK::KeyFinder.new(jwks: @options[:jwks]).key_for(header['kid']) if @options[:jwks]
+      if (x5c_options = @options[:x5c])
+        @key = X5cKeyFinder.new(x5c_options[:root_certificates], x5c_options[:crls]).from(header['x5c'])
+      end
     end
 
     def verify_signature_for?(key)
--- a/lib/jwt/x5c_key_finder.rb
+++ b/lib/jwt/x5c_key_finder.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+require 'base64'
+require 'jwt/error'
+
+module JWT
+  # If the x5c header certificate chain can be validated by trusted root
+  # certificates, and none of the certificates are revoked, returns the public
+  # key from the first certificate.
+  # See https://tools.ietf.org/html/rfc7515#section-4.1.6
+  class X5cKeyFinder
+    def initialize(root_certificates, crls = nil)
+      raise(ArgumentError, 'Root certificates must be specified') unless root_certificates
+
+      @store = build_store(root_certificates, crls)
+    end
+
+    def from(x5c_header_or_certificates)
+      signing_certificate, *certificate_chain = parse_certificates(x5c_header_or_certificates)
+      store_context = OpenSSL::X509::StoreContext.new(@store, signing_certificate, certificate_chain)
+
+      if store_context.verify
+        signing_certificate.public_key
+      else
+        error = "Certificate verification failed: #{store_context.error_string}."
+        if (current_cert = store_context.current_cert)
+          error = "#{error} Certificate subject: #{current_cert.subject}."
+        end
+
+        raise(JWT::VerificationError, error)
+      end
+    end
+
+    private
+
+    def build_store(root_certificates, crls)
+      store = OpenSSL::X509::Store.new
+      store.purpose = OpenSSL::X509::PURPOSE_ANY
+      store.flags = OpenSSL::X509::V_FLAG_CRL_CHECK | OpenSSL::X509::V_FLAG_CRL_CHECK_ALL
+      root_certificates.each { |certificate| store.add_cert(certificate) }
+      crls&.each { |crl| store.add_crl(crl) }
+      store
+    end
+
+    def parse_certificates(x5c_header_or_certificates)
+      if x5c_header_or_certificates.all? { |obj| obj.is_a?(OpenSSL::X509::Certificate) }
+        x5c_header_or_certificates
+      else
+        x5c_header_or_certificates.map do |encoded|
+          OpenSSL::X509::Certificate.new(::Base64.strict_decode64(encoded))
+        end
+      end
+    end
+  end
+end
--- a/spec/x5c_key_finder_spec.rb
+++ b/spec/x5c_key_finder_spec.rb
@@ -1,0 +1,203 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'jwt/x5c_key_finder'
+
+describe JWT::X5cKeyFinder do
+  let(:root_key) { OpenSSL::PKey.read(File.read(File.join(CERT_PATH, 'rsa-2048-private.pem'))) }
+  let(:root_dn) { OpenSSL::X509::Name.parse('/DC=org/DC=fake-ca/CN=Fake CA') }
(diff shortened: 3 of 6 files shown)


rahulbajaj0509:
@bdewater i am very naive at this and want to test this PR. Would you mind explaining me  what CRL is and some detailed steps that would help me to test the PR?

bdewater:
@rahulbajaj0509 a [Certificate Revocation List](https://tools.ietf.org/html/rfc5280#section-5) is a list of certificates that should be considered invalid, even if the signature is correct, not before/not after timestamps are valid, etc. The list contains a reasonCode as to why a certificate was revoked. 

My reason for opening this PR was because I missed this functionality for verifying the FIDO Alliance metadata files as part of https://github.com/cedarcode/webauthn-ruby/pull/208 - I haven't had the chance yet to test this PR within the context of that work, but I'm planning to do that soon and make any adjustments if needed.

You can find out where to obtain the CRL file by seeing if any certificate (either part of the x5c header or the supplied root certificates) has a CRL Distribution Points extension. [A helper method in the OpenSSL gem](https://github.com/ruby/openssl/pull/275) to get CRL Distribution Points URIs was merged and is slated for the upcoming release. That PR builds on other unreleased work, so until the next OpenSSL gem release (hopefully soon) you can use this:
```ruby
def crl_uris(certificate)
  ext = certificate.extensions.find { |ext| ext.oid == "crlDistributionPoints" }
  return nil if ext.nil?

  ext_value_der = OpenSSL::ASN1.decode(ext.to_der).value.last.value
  cdp_asn1 = OpenSSL::ASN1.decode(ext_value_der)
  if cdp_asn1.tag_class != :UNIVERSAL || cdp_asn1.tag != OpenSSL::ASN1::SEQUENCE
    raise OpenSSL::ASN1::ASN1Error "invalid extension"
  end

  crl_uris = cdp_asn1.map do |crl_distribution_point|
    distribution_point = crl_distribution_point.value.find do |v|
      v.tag_class == :CONTEXT_SPECIFIC && v.tag == 0
    end
    full_name = distribution_point&.value&.find do |v|
      v.tag_class == :CONTEXT_SPECIFIC && v.tag == 0
    end
    full_name&.value&.find do |v|
      v.tag_class == :CONTEXT_SPECIFIC && v.tag == 6 # uniformResourceIdentifier
    end
  end

  crl_uris&.map(&:value)
end
```

Let's take the [example x5c header](https://tools.ietf.org/html/rfc7515#appendix-B) from the RFC. Once you have base64 decoded these, you'll see these certificates have been issued by GoDaddy. We'll assume here that you are certain whomever will be giving you JWTs will use GoDaddy as their root CA to sign their intermediate certificates, which will in turn be used to sign the JWTs.

Head over to https://certs.godaddy.com/repository to obtain the GoDaddy root certificates. Use the `crl_uris` method to extract the CRL URIs and download those files ([1](http://certificates.godaddy.com/repository/gdroot.crl), [2](http://certificates.godaddy.com/repository/root.crl)) as well. Then you should be able to call 

```ruby
root_certificates = [] # Array of OpenSSL::X509::Certificate objects
crls = [] # Array of OpenSSL::X509::CRL objects
JWT.decode(jwt, nil, true, root_certificates: root_certificates, crls: crls)
```

cpb:
This looks like a great enhancement. If I get some time over the long weekend, how could a first-time-contributor to ruby-jwt help move this forward?

bdewater:
@cpb if you can use this fork in your app and provide feedback if it works for you or have any other comments, that'd be great. I haven't had the time yet to do that myself.

bdewater:
I verified this code works for my use case (verifying WebAuthn attestation via the FIDO Alliance Metadata Service). Code can be seen here: https://github.com/bdewater/fido_metadata/commit/b157b5aedf912ad7f2913439bca807ab83b73ee4 - any feedback welcome 🙂

spec/x5c_key_finder_spec.rb
```diff
@@ -0,0 +1,174 @@
...
+
+  def issue_crl(revocations, last_up: nil, next_up: nil, issuer:, issuer_key:)
+    crl = OpenSSL::X509::CRL.new
+    crl.issuer = issuer.subject
+    crl.version = 1
+    now = Time.now
+    crl.last_update = last_up || now - 3600
+    crl.next_update = next_up || now + 3600
+
+    revocations.each do |rserial, time, reason_code|
```
thiennguyenmark, line 158: better getting rebase master herer

lib/jwt/decode.rb
```diff
@@ -32,12 +34,15 @@ def decode_segments
 
     private
 
+    def verify_algo
```
anakinj, line 37: This was a nice extraction!

lib/jwt/decode.rb
```diff
@@ -32,12 +34,15 @@ def decode_segments
...
+      raise(JWT::IncorrectAlgorithm, 'Expected a different algorithm') unless options_includes_algo_in_header?
+    end
+
     def verify_signature
       @key = find_key(&@keyfinder) if @keyfinder
       @key = ::JWT::JWK::KeyFinder.new(jwks: @options[:jwks]).key_for(header['kid']) if @options[:jwks]
-
-      raise(JWT::IncorrectAlgorithm, 'An algorithm must be specified') if allowed_algorithms.empty?
-      raise(JWT::IncorrectAlgorithm, 'Expected a different algorithm') unless options_includes_algo_in_header?
+      @key = JWT::X5cKeyFinder.from(header['x5c'], @options.fetch(:root_certificates), @options[:crls]) if header['x5c']
```
anakinj, line 45: Think the trigger for this needs to be indicated by the application requesting the verification. Now it seems to be only the `x5c` header that controls when this happens.

Also a test for this path would be valuable.

Maybe something like this could trigger the x5c logic.

```
JWT.decode(token, nil, true, x5c: {root_certificates: [], crls: [] })
```

anakinj, line 45: Is there a reason why `@options.fetch(:root_certificates)` and not `@options[:root_certificates]`?

lib/jwt/x5c_key_finder.rb
```diff
@@ -0,0 +1,48 @@
...
+    def self.parse_certificates(x5c_header_or_certificates)
+      if x5c_header_or_certificates.all? { |obj| obj.is_a?(OpenSSL::X509::Certificate) }
+        x5c_header_or_certificates
+      else
+        x5c_header_or_certificates.map do |encoded|
+          OpenSSL::X509::Certificate.new(::Base64.strict_decode64(encoded))
+        end
+      end
+    end
+    private_class_method :parse_certificates
```
anakinj, line 36: Instead of making the static methods private afterwards with `private_class_method` you could use`class << self` combined with a `private` block.

```
class X5cKeyFinder
  class << self
    def from(x5c_header_or_certificates, root_certificates, crls)
    end
    
    private

    def parse_certificates(x5c_header_or_certificates)
    end

    def build_store(root_certificates, crls)
    end
  end
end

```

anakinj:
Hi, this looks really promising 👍. Gave a few comments here and there, hope it's of any value. A little something for the documentation would also be great.

thomasdarde:
Any update ont his PR Review ?

spiffistan:
Any chance this can get merged?

lib/jwt/decode.rb
bdewater: A `nil` value for root certificates doesn't make sense - it's not an optional value unlike CRLs. Raising an error here is more clear it's being used wrong than in `X5cKeyFinder.build_store`.

bdewater: Refactored this to happen inside the keyfinder class.

bdewater:
@anakinj I rebased and implemented your feedback. Curious how strong you feel about the remaining Reek comments - shoveling these private functions into the SecurityUtils module and/or breaking them up to lower statement count won't add much real value IMO.

anakinj:
I agree with you that the suggestion of the robot should not be taken as the truth. I will take a look at this when I have a chance to fully concentrate on it. Too complicated after a long day :)

lib/jwt/x5c_key_finder.rb
```diff
@@ -0,0 +1,54 @@
...
+require 'jwt/error'
+
+module JWT
+  # If the x5c header certificate chain can be validated by trusted root
+  # certificates, and none of the certificates are revoked, returns the public
+  # key from the first certificate.
+  # See https://tools.ietf.org/html/rfc7515#section-4.1.6
+  class X5cKeyFinder
+    def initialize(root_certificates, crls = nil)
+      raise(JWT::DecodeError, 'Root certificates must be specified') unless root_certificates
```
anakinj, line 13: Could as well just be a ArgumentError

lib/jwt/x5c_key_finder.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    private
+
+    def build_store(root_certificates, crls)
+      store = OpenSSL::X509::Store.new
+      store.purpose = OpenSSL::X509::PURPOSE_ANY
+      store.flags = OpenSSL::X509::V_FLAG_CRL_CHECK | OpenSSL::X509::V_FLAG_CRL_CHECK_ALL
+      root_certificates.each { |certificate| store.add_cert(certificate) }
+      crls && crls.each { |crl| store.add_crl(crl) }
```
anakinj, line 40: An alternative way to ensure the array is iterateable would be `Array(crls).each {|crl|}`

anakinj (approved):
Looking good.

anakinj:
@bdewater Is this ready to get merged?

bdewater:
@anakinj it is now 😄 rebased to pick up the latest Rubocop stuff, addressed your most recent feedback, and added a test to ensure the `x5c` option in the `JWT.decode` call will actually call the `X5cKeyFinder` class.

anakinj (approved):

anakinj:
Thank you for your contribution (and patience) @bdewater.

Issue #95, opened 2015-08-05
Options hash uses both symbols and strings as keys.


aj-michael:
I tried to verify the issuer on a JWT that I was decoding today, and I found this line (and others like it) very ugly.

```
if options[:verify_iss] && options['iss']
```

This means that my code to decode the JWT looked like this:

```
JWT.decode(token, public_key, true, :verify_iss => true, 'iss' => 'my issuer')
```

In my opinion, the consumer of this API should be be able to do one of the following:

1. Always use symbols as keys in the options hash.
2. Always use strings as keys in the options hash.
3. Be able to either use symbols or keys in the options in hash and have our library deal with it under the hood.

Thoughts?

yordis:
:+1: for symbol only until someone complain about it :smile:

excpt:
:+1: symbols only :)

skippy:
+1

This is an issue beyond semantics... @aj-michael  it is a security hole.

I pass in the following:
`{:aud=>"test_aud_id", :verify_aud=>true}`

and it decodes the token successfully.  The hole is it decodes *regardless* of what the `aud` is.  You can see this bug in line jwt.rb#172 `if options[:verify_aud] && options['aud']`

if I pass in:
`{:aud=>"test_aud_id", :verify_aud=>true, :verify_iat=>true, :sub=>"test_sub_id", :verify_sub=>true}`

this fails because how the `verify_sub` checks, it doesn't allow it to silently fail (even though `sub` is set in the token)

excpt:
@aj-michael @yordis @skippy :)

skippy:
@excpt fair enough!  I wasn't trying to be snarky.  Thank you for the fast PR!

yordis:
@excpt awesome, very good clean there :+1:

PR #348, opened 2020-03-25
Handle parsed JSON JWKS input with string keys

martinemde:
When decoded JSON is passed to the JWK.import, there are difficult to diagnose failures that occur. 

Since the goal of JWKS is to support loading of json files, we should cleanly support string keys as that is the natural input to these methods.

This PR creates support for directly parsed JSON with string keys.
* No realistic performance decrease.
* Backwards compatibility is maintained by searching for symbol keys and then string keys.
* Improved error messaging that would have helped me find this issue more quickly.

I have copied below the spec output without the fixes I added. You can see from the failures that the errors are very difficult to understand when all that has happened is the keys were missed.

For instance, "could not find public key" is confusing because debugging shows a perfect match between your JWK and the JWT kid. It's simply the fact that the key is a string and this is not expected.

1) JWT.decode for JWK usecase when jwk keys are loaded from JSON with string keys decodes the token
     Failure/Error: raise ::JWT::DecodeError, "Could not find public key for kid #{kid}" unless jwk

```ruby
JWT::DecodeError:
  Could not find public key for kid 3240eb23aa4f72d26b1b1b534fee4c58c68a51af73b18acedd1e697a775a67fe
# ./lib/jwt/jwk/key_finder.rb:17:in `key_for'
# ./lib/jwt/decode.rb:37:in `verify_signature'
# ./lib/jwt/decode.rb:26:in `decode_segments'
# ./lib/jwt.rb:28:in `decode'
# ./spec/jwk/decode_with_jwk_spec.rb:71:in `block (4 levels) in <top (required)>'
```

  2) JWT::JWK::RSA.import when jwk_data is given without e and/or n raises an error
     Failure/Error: expect { subject }.to raise_error(JWT::JWKError, "Key format is invalid for RSA")

```ruby
expected JWT::JWKError with "Key format is invalid for RSA", got #<NoMethodError: undefined method `tr' for nil:NilClass> with backtrace:
# ./lib/jwt/jwk/rsa.rb:43:in `import'
# ./spec/jwk/rsa_spec.rb:59:in `block (3 levels) in <top (required)>'
# ./spec/jwk/rsa_spec.rb:83:in `block (5 levels) in <top (required)>'
# ./spec/jwk/rsa_spec.rb:83:in `block (4 levels) in <top (required)>'
# ./spec/jwk/rsa_spec.rb:83:in `block (4 levels) in <top (required)>'
```

  3) JWT::JWK::RSA.import when keypair is imported with string keys from JSON returns a hash with the public parts of the key

```ruby
Failure/Error:
  imported_key.set_key(OpenSSL::BN.new(::Base64.urlsafe_decode64(jwk_data[:n]), BINARY),
    OpenSSL::BN.new(::Base64.urlsafe_decode64(jwk_data[:e]), BINARY),
    nil)

NoMethodError:
  undefined method `tr' for nil:NilClass
# ./lib/jwt/jwk/rsa.rb:43:in `import'
# ./spec/jwk/rsa_spec.rb:59:in `block (3 levels) in <top (required)>'
# ./spec/jwk/rsa_spec.rb:74:in `block (4 levels) in <top (required)>'
```

Linking to #95 also because it is somewhat related to the string/symbol issue here.

EDIT:

The source of this problem could be tied back to the README example which uses `.export` instead of parsing an actual parsed JSON Web Key as the input to the key finder. We could update `export` to return string keys to make it more clear.

### DIFF
--- a/lib/jwt/jwk.rb
+++ b/lib/jwt/jwk.rb
@@ -12,9 +12,10 @@ module JWT
 
     class << self
       def import(jwk_data)
-        raise JWT::JWKError, 'Key type (kty) not provided' unless jwk_data[:kty]
+        jwk_kty = jwk_data[:kty] || jwk_data['kty']
+        raise JWT::JWKError, 'Key type (kty) not provided' unless jwk_kty
 
-        MAPPINGS.fetch(jwk_data[:kty].to_s) do |kty|
+        MAPPINGS.fetch(jwk_kty.to_s) do |kty|
           raise JWT::JWKError, "Key type #{kty} not supported"
         end.import(jwk_data)
       end
--- a/lib/jwt/jwk/key_finder.rb
+++ b/lib/jwt/jwk/key_finder.rb
@@ -14,10 +14,8 @@ module JWT
 
         jwk = resolve_key(kid)
 
-        unless jwk
-          raise ::JWT::DecodeError, 'No keys found in jwks' if jwks_keys.empty?
-          raise ::JWT::DecodeError, "Could not find public key for kid #{kid}"
-        end
+        raise ::JWT::DecodeError, 'No keys found in jwks' if jwks_keys.empty?
+        raise ::JWT::DecodeError, "Could not find public key for kid #{kid}" unless jwk
 
         ::JWT::JWK.import(jwk).keypair
       end
--- a/lib/jwt/jwk/key_finder.rb
+++ b/lib/jwt/jwk/key_finder.rb
@@ -14,7 +14,10 @@ module JWT
 
         jwk = resolve_key(kid)
 
-        raise ::JWT::DecodeError, "Could not find public key for kid #{kid}" unless jwk
+        unless jwk
+          raise ::JWT::DecodeError, "No keys found in jwks" if jwks_keys.empty?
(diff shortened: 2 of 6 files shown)


anakinj:
Hi, this feature is great. It's always a hassle to ensure the keys are the correct type.

Im wondering if it would be a good idea to transform the keys into either strings or symbols before using the hash, that would remove the need to always check for both key types whenever the values are accessed.

martinemde:
> Im wondering if it would be a good idea to transform the keys into either strings or symbols before using the hash, that would remove the need to always check for both key types whenever the values are accessed.

I think most of the hash lookups are during import so it wouldn't buy us much.

The [example in the readme for loading JWK](https://github.com/jwt/ruby-jwt#json-web-key-jwk) indicates that caching should be performed inside the loader (before this import step). Unless I'm mistaken, after importing, a single key is grabbed from the key set and used to check the signature, then the entire thing is discarded.

Given just how very fast 2 hash lookups is compared to deep copying an parsed json hash of unknown size, I'd recommend we don't convert the input in any way.

martinemde:
Yay, thank you!

excpt:
> Yay, thank you!

I thank you for the contribution! Thanks! 👍

PR #349, opened 2020-03-25
Hash jwks keys by kid

martinemde:
Depends on #348.

I found this improvement when trying to make the code better for fetching keys by kid. You can see it's a little messy in my other branch and cleaner here.

I separated it into another PR since this refactor changes things unrelated to parsing string keys.


martinemde:
As discussed in #348, I realize this keyloader is short-lived and so the optimization here does nothing. Closing.

PR #352, opened 2020-04-28
Improve support for Unsecured JWT

### LINKED ISSUE #323: When using none alg (no signature), should the claims verification be ignored?
Hi,

When I upgrade from v1.5.6 to v2.2.1, I find that, the 'none' algorithms behaves differently.

In v1.5.6, even use 'none', the exp claim would be verified, but in v2.2.1, it doesn't

the script:
```ruby
payload = { 'exp' => (Time.now.to_i - 5) }
id_token = JWT.encode(payload, nil, 'none')
JWT.decode(id_token, nil, false)
```
the output:
- v1.5.6: 
```ruby
=> JWT::ExpiredSignature: Signature has expired
```
- v2.2.1: 
```ruby
=> [{"exp"=>1561080813}, {"alg"=>"none"}]
```

And I find that this commit https://github.com/jwt/ruby-jwt/commit/67f4a5ad25ec2695b1dc414a8797bd5a0deaaec4 change the behavior.

I want to make sure that, when using none alg (no signature), should the claims verification be ignored?

Thank you!

hesalx:
This allows for symmetric support of Unsecured JWT. It now works as any other algorithm for both encoding and decoding.

Fixes #323.

https://tools.ietf.org/html/rfc7519#section-6 allows for Unsecured JWT using "alg" value of "none". It does not forbid further processing of Unsecured JWT and thus claim verification still applies.

https://tools.ietf.org/html/rfc7518#section-3.6 requires the signature to be an empty string.

https://tools.ietf.org/html/rfc7518#section-3.6 and https://tools.ietf.org/html/rfc7518#section-8.5 forbid accepting Unsecured JWT unless explicitly allowed (this has already been addressed).

While not explicitly addressed by any RFC, based on common sense this commit explicitly forbids the use of any signing key for Unsecured JWT. As Unsecured JWT requires an empty signature and cannot possibly make use of a key, it should be considered an error to use a key together with algorithm value of "none". For an Unsecured JWT the signing key should be set to either false or nil.

### DIFF
--- a/lib/jwt/decode.rb
+++ b/lib/jwt/decode.rb
@@ -13,7 +13,7 @@ module JWT
       @jwt = jwt
       @key = key
       @options = options
-      @segments = jwt.split('.')
+      @segments = jwt.split('.', -1)
       @verify = verify
       @signature = ''
       @keyfinder = keyfinder
@@ -66,7 +66,6 @@ module JWT
 
     def validate_segment_count!
       return if segment_length == 3
-      return if !@verify && segment_length == 2 # If no verifying required, the signature is not needed
 
       raise(JWT::DecodeError, 'Not enough or too many segments')
     end
--- a/spec/jwt_spec.rb
+++ b/spec/jwt_spec.rb
@@ -44,6 +44,7 @@ describe JWT do
 
   context 'alg: NONE' do
     let(:alg) { 'none' }
+    let(:sig) { 'kWOVtIOpWcG7JnyJG0qOkTDbOy636XrrQhMm_8JrRQ8' }
 
     it 'should generate a valid token' do
       token = JWT.encode payload, nil, alg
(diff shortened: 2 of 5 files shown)


lib/jwt/decode.rb
```diff
@@ -13,7 +13,7 @@ def initialize(jwt, key, verify, options, &keyfinder)
       @jwt = jwt
       @key = key
       @options = options
-      @segments = jwt.split('.')
+      @segments = jwt.split('.', -1)
```
hesalx, line 16: `#split` by default ignores empty strings. This change is needed to correctly identify an empty string signature in a JWT structured according to https://tools.ietf.org/html/rfc7519#section-7.2 and https://tools.ietf.org/html/rfc7516#section-9

lib/jwt/decode.rb
```diff
@@ -66,7 +66,6 @@ def verify_claims
 
     def validate_segment_count!
       return if segment_length == 3
-      return if !@verify && segment_length == 2 # If no verifying required, the signature is not needed
```
hesalx, line 69: It seems that in no point any RFC allows for a 2-segment JWT.

Very likely this line was there only sue to how `#split` works by default.

All relevant RFCs explicitly define JWT to have two dots (with the possibility of the last segment, the signature, to be an empty string).
See https://tools.ietf.org/html/rfc7515#section-3.1 and https://tools.ietf.org/html/rfc7516#section-9

The "at least one period ('.')" in https://tools.ietf.org/html/rfc7519#section-7.2 does not actually signify that there exists any valid JWT format with "exactly one period".

spec/jwt_spec.rb
```diff
@@ -358,12 +384,6 @@
     end
   end
 
-  context 'a token with two segments but does not require verifying' do
-    it 'raises something else than "Not enough or too many segments"' do
-      expect { JWT.decode('ThisIsNotAValidJWTToken.second', nil, false) }.to raise_error(JWT::DecodeError, 'Invalid segment encoding')
-    end
-  end
-
```
hesalx, line 366: See the above comments regarding segments count.

There existed circumstances where 2-segment token would be processed without an exception (`"alg": "none"` along with `verify=false`) while such a format is not defined by any RFC and should be invalid.

If this is considered a breaking change we may handle it separately from this PR.

glasses618:
Hi @hesalx , just want to remind you that. If you put the issue link in the commit message. Every time you force-pushed will add a message automatically in the original issue page this
image
Maybe you could remove the link from your commit message if this is not the behavior you expect.

lib/jwt/decode.rb
excpt: This would introduce a breaking change.

Experiences from past breaking changes showed that this kind of change creates a lot of chaos out there.

I recommend creating a separate PR.

excpt: On second thoughts. The whole PR will be a breaking change then. So no separate PR needed. Scheduled for 3.0.0 milestone.

anakinj:
@hesalx Are you still active in the Ruby world? I'm thinking we could include the suggested improvements to the none algorithm in a 3.0 version of the gem.

PR #356, opened 2020-07-02
Pin simplecov to < 0.18

anakinj:
Something stopped being compatible with simplecov 0.18 so pinning the version for now.

Think we need to do something about all the coverage reporting dependencies, are they adding value?


anakinj:
@excpt Think we could have this in master for the CI to pass. Then the pending PRs would need a rebase to get greener CIs

PR #368, opened 2020-08-29
Add Truffleruby head to CI

gogainda:


gogainda:
@excpt tests are green, pls review/merge

excpt (approved):

excpt:
@gogainda Thanks for the contribution. 👍

gogainda:
@eregon

anakinj:
@gogainda It feels like this truffleruby is making the CI fail randomly, for example today the HEAD package link is broken and this happens: https://travis-ci.org/github/jwt/ruby-jwt/jobs/734205049

Should this be pinned to a specific truffleruby version and not to head?

gogainda:
@eregon any ideas why it may happen?

eregon:
That seems an issue of RVM, curl, or GitHub, but not of TruffleRuby itself.
It seems TravisCI hits some limit, and `RVM` or `curl` doesn't seem to retry: https://github.com/rvm/rvm/issues/4992
Feel free to add TruffleRuby in `allow_failures` until this is solved.

eregon:
BTW, I don't think that link was ever broken, but the error seen by TravisCI is:
```
curl: (22) The requested URL returned error: 429 too many requests
```

eregon:
One way to workaround this is to use GitHub Actions, that's also significantly faster and more reliable than TravisCI in my experience. But of course migrating to it is some effort.

PR #369, opened 2020-09-08
Add support for JWKs with EC key type

richardlarocque:
Adds support for JWKs with "kty" value "EC".  This commit includes
support for curves ("crv") "P-256", "P-384", "P-521".

For additional details on these JWKs and their contents, see
https://tools.ietf.org/html/rfc7518#section-6.2.1.

This implementation adheres closely to the pattern set by
`JWT::JWK::RSA`.  It keeps the same coding style and method names.
It also inherits a few quirks:
- It ignores any private key ("d") values when importing JWKs.
- It never emits private key ("d") values when exporting JWKs.
- An import followed by an export does not preserve the "kid" value.

These behaviors seem strange to me, but I worry that changing the
convention would be surprising and potentially dangerous to existing
users of this library.

### DIFF
--- a/lib/jwt/jwk.rb
+++ b/lib/jwt/jwk.rb
@@ -1,13 +1,16 @@
 # frozen_string_literal: true
 
 require_relative 'jwk/rsa'
+require_relative 'jwk/ec'
 require_relative 'jwk/key_finder'
 
 module JWT
   module JWK
     MAPPINGS = {
       'RSA' => ::JWT::JWK::RSA,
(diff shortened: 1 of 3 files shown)


richardlarocque:
Sorry, this was not intended for upstreaming just yet.  I opened this PR against the wrong repo.  I was planning to try this out in a forked repo before opening a PR here.

I'm going to close this for now.  Sorry for the noise.  If all goes well, I'll be back with a better PR soon.

PR #371, opened 2020-09-18
Add support for JWKs with EC key type

richardlarocque:
Adds support for JWKs with "kty" value "EC".  This commit includes
support for curves ("crv") "P-256", "P-384", "P-521".

For additional details on these JWKs and their contents, see
https://tools.ietf.org/html/rfc7518#section-6.2.1.

This implementation adheres closely to the pattern set by
`JWT::JWK::RSA`.  It keeps the same coding style and method names.
It also inherits a few quirks:
- It ignores any private key ("d") values when importing JWKs.
- It never emits private key ("d") values when exporting JWKs.
- An import followed by an export does not preserve the "kid" value.

These behaviors seem strange to me, but I worry that changing the
convention would be surprising and potentially dangerous to existing
users of this library.

### DIFF
--- a/lib/jwt/jwk/ec.rb
+++ b/lib/jwt/jwk/ec.rb
@@ -7,21 +7,20 @@ module JWT
       def_delegators :@keypair, :private?, :public_key
 
       attr_reader :keypair
+      attr_reader :kid
 
       KTY    = 'EC'.freeze
       BINARY = 2
 
-      def initialize(keypair)
+      def initialize(keypair, kid = nil)
         raise ArgumentError, 'keypair must be of type OpenSSL::PKey::EC' unless keypair.is_a?(OpenSSL::PKey::EC)
 
         @keypair = keypair
+        @kid = kid || generate_kid(@keypair)
       end
 
       def export
-        crv, x_octets, y_octets = keypair_components
-        sequence = OpenSSL::ASN1::Sequence([OpenSSL::ASN1::Integer.new(OpenSSL::BN.new(x_octets, BINARY)),
-                                            OpenSSL::ASN1::Integer.new(OpenSSL::BN.new(y_octets, BINARY))])
-        kid = OpenSSL::Digest::SHA256.hexdigest(sequence.to_der)
+        crv, x_octets, y_octets = keypair_components(keypair)
         {
           kty: KTY,
           crv: crv,
@@ -33,9 +32,16 @@ module JWT
 
       private
 
-      def keypair_components
-        encoded_point = keypair.public_key.to_bn.to_s(BINARY)
-        case keypair.group.curve_name
+      def generate_kid(ec_keypair)
+        _crv, x_octets, y_octets = keypair_components(ec_keypair)
+        sequence = OpenSSL::ASN1::Sequence([OpenSSL::ASN1::Integer.new(OpenSSL::BN.new(x_octets, BINARY)),
+                                            OpenSSL::ASN1::Integer.new(OpenSSL::BN.new(y_octets, BINARY))])
+        OpenSSL::Digest::SHA256.hexdigest(sequence.to_der)
+      end
+
+      def keypair_components(ec_keypair)
+        encoded_point = ec_keypair.public_key.to_bn.to_s(BINARY)
+        case ec_keypair.group.curve_name
         when 'prime256v1'
           crv = 'P-256'
           x_octets, y_octets = encoded_point.unpack('xa32a32')
@@ -46,7 +52,7 @@ module JWT
           crv = 'P-521'
           x_octets, y_octets = encoded_point.unpack('xa66a66')
         else
-          raise "Unsupported curve '#{keypair.group.curve_name}'"
+          raise "Unsupported curve '#{ec_keypair.group.curve_name}'"
         end
         [crv, x_octets, y_octets]
       end
@@ -64,10 +70,10 @@ module JWT
           # See https://tools.ietf.org/html/rfc7518#section-6.2.1 for an
           # explanation of the relevant parameters.
 
-          jwk_crv, jwk_x, jwk_y = jwk_attrs(jwk_data, %i[crv x y])
+          jwk_crv, jwk_x, jwk_y, jwk_kid = jwk_attrs(jwk_data, %i[crv x y kid])
           raise Jwt::JWKError, 'Key format is invalid for EC' unless jwk_crv && jwk_x && jwk_y
 
-          new(ec_pkey(jwk_crv, jwk_x, jwk_y))
+          new(ec_pkey(jwk_crv, jwk_x, jwk_y), jwk_kid)
         end
 
         def to_openssl_curve(crv)
--- a/lib/jwt/jwk/ec.rb
+++ b/lib/jwt/jwk/ec.rb
@@ -78,7 +78,7 @@ module JWT
           when 'P-256' then 'prime256v1'
           when 'P-384' then 'secp384r1'
           when 'P-521' then 'secp521r1'
-          else raise 'Invalid curve provided'
+          else raise JWT::JWKError, 'Invalid curve provided'
           end
         end
 
--- a/lib/jwt/jwk/ec.rb
+++ b/lib/jwt/jwk/ec.rb
@@ -1,0 +1,129 @@
+# frozen_string_literal: true
+
+module JWT
+  module JWK
+    class EC
+      extend Forwardable
+      def_delegators :@keypair, :private?, :public_key
+
+      attr_reader :keypair
+
+      KTY    = 'EC'.freeze
+      BINARY = 2
+
+      def initialize(keypair)
+        raise ArgumentError, 'keypair must be of type OpenSSL::PKey::EC' unless keypair.is_a?(OpenSSL::PKey::EC)
+
+        @keypair = keypair
+      end
+
+      def export
+        crv, x_octets, y_octets = keypair_components
+        sequence = OpenSSL::ASN1::Sequence([OpenSSL::ASN1::Integer.new(OpenSSL::BN.new(x_octets, BINARY)),
+                                            OpenSSL::ASN1::Integer.new(OpenSSL::BN.new(y_octets, BINARY))])
+        kid = OpenSSL::Digest::SHA256.hexdigest(sequence.to_der)
+        {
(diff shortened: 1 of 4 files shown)


anakinj:
Looks great, thanks for your contribution.

Wondering about this 'An import followed by an export does not preserve the "kid" value.', this behaviour was not intended  and is fixed in #320 for the RSA keys. Maybe this feature could be perfect from the beginning in regards to that :)

About not exporting private keys, I recall me thinking that it could be dangerous to export the private key by default. If we would like to support this we could add a parameter to the export method that would allow inclusion of the private parts.

lib/jwt/jwk/ec.rb
```diff
@@ -0,0 +1,129 @@
...
+
+        def to_openssl_curve(crv)
+          # The JWK specs and OpenSSL use different names for the same curves.
+          # See https://tools.ietf.org/html/rfc5480#section-2.1.1.1 for some
+          # pointers on different names for common curves.
+          case crv
+          when 'P-256' then 'prime256v1'
+          when 'P-384' then 'secp384r1'
+          when 'P-521' then 'secp521r1'
+          else raise 'Invalid curve provided'
```
anakinj, line 81: Maybe we could come up with a type for the exceptions without types?

anakinj: There is the `::JWT::JWKError` that would at least be generic enough? or then introduce something new inherited from that?

lib/jwt/jwk/ec.rb
```diff
@@ -0,0 +1,135 @@
...
+          crv = 'P-256'
+          x_octets, y_octets = encoded_point.unpack('xa32a32')
+        when 'secp384r1'
+          crv = 'P-384'
+          x_octets, y_octets = encoded_point.unpack('xa48a48')
+        when 'secp521r1'
+          crv = 'P-521'
+          x_octets, y_octets = encoded_point.unpack('xa66a66')
+        else
+          raise "Unsupported curve '#{ec_keypair.group.curve_name}'"
```
anakinj, line 55: Missed this the first time around. The `JWT::JWKError` could be used here also

richardlarocque: My bad.  Should be fixed now.

phlegx:
Hi! Why `JWT::JWK::RSA` and `JWT::JWK::EC` don't export private keys? It is defined in [RFC7517](https://tools.ietf.org/html/rfc7517#appendix-A) that also private keys can be exported. Also other libraries [do that](https://github.com/nov/json-jwt/blob/master/lib/json/jwk/jwkizable.rb).

anakinj:
Hi @phlegx 

Think that it would be a nice feature.

As mentioned before the main reason for it was to avoid accidental leaking of the private key, so still think that including the private key in the export should not be the default behaviour. In other words the private key should only be included when asked for.

anakinj:
@richardlarocque Looks good.

anakinj (approved):

phlegx:
@anakinj the RFC describes JWK with:

_A JSON Web Key (JWK) is a JavaScript Object Notation (JSON) data structure that represents a cryptographic key._

So, JWK is only a data structure that represents a cryptographic key. How a user expose a JWK to a web endpoint is not part of the [RFC7517](https://tools.ietf.org/html/rfc7517) and should not affect the behavior or structure of a JWK.

anakinj:
The need until now has been focused on the public parts of the keys (signature verification and exports in tests to simulate public JWK endpoints). So therefore the private parts have been totally left out.

Think that if we would support exporting of the private keys we should protect the uneducated user from leaking sensitive information and require the user to actually ask for the private parts to be included.

Made  a quick proposal on how the private parts of the key could be included in the export #373

phlegx:
@anakinj it is clear for me thx! It is a good compromise for me to add an option if private key should be exported or not. We should focus to add this feature to RSA, EC and HMAC and if possible release a new version.

phlegx:
@richardlarocque I suggest to merge master branch to your PR, that includes the abstract class. Then you can adapt your PR to the abstract structure and the `include_private` option.

richardlarocque:
Merged in upstream changes to fix conflicts.  Please take another look and merge if desired.  (I don't have permission to hit the "merge" button, even if there is an approving review.)

lib/jwt/jwk/ec.rb
```diff
@@ -0,0 +1,135 @@
+# frozen_string_literal: true
+
+module JWT
+  module JWK
+    class EC
```
phlegx, line 5: You have to inherit from abstract class here:

`class EC < KeyAbstract`

lib/jwt/jwk/ec.rb
```diff
@@ -0,0 +1,135 @@
...
+      extend Forwardable
+      def_delegators :@keypair, :private?, :public_key
+
+      attr_reader :keypair
+      attr_reader :kid
```
phlegx, line 10: This can be removed, because already included from abstract class.

lib/jwt/jwk/ec.rb
```diff
@@ -0,0 +1,135 @@
...
+
+      KTY    = 'EC'.freeze
+      BINARY = 2
+
+      def initialize(keypair, kid = nil)
+        raise ArgumentError, 'keypair must be of type OpenSSL::PKey::EC' unless keypair.is_a?(OpenSSL::PKey::EC)
+
+        @keypair = keypair
```
phlegx, line 18: Remove this line and add `super` instead.

lib/jwt/jwk/ec.rb
```diff
@@ -0,0 +1,135 @@
...
+        @kid = kid || generate_kid(@keypair)
+      end
+
+      def export
```
phlegx, line 22: Add parameter `options` like: `def export(options = {})`

lib/jwt/jwk/ec.rb
```diff
@@ -0,0 +1,135 @@
...
+        crv, x_octets, y_octets = keypair_components(keypair)
+        {
```
phlegx, line 24: Export private key if option `include_private` is set to true like HMAC JWK does: https://github.com/jwt/ruby-jwt/blob/c5bf3376a6ff8bd619bd724815d9e8e6bfc3accd/lib/jwt/jwk/hmac.rb#L30

phlegx (commented):
Can you please implement methods `private?` and `public_key` like JWK HMAC and RSA (PR #373) does?

excpt (approved):

richardlarocque:
Sorry, I got distracted with other things and ended up not paying much attention to this PR.

I find that abstract classes are not very common in Ruby.  There's no need to define stub methods that throw exceptions when they're not overridden; that's what `NoMethodError` is for.  Ruby provides no compile-time enforcement of interfaces, so the best way to prove that two classes implement the same interface is with generic unit tests.  If you need to share functionality between classes, that's usually done with mixins.  In other words, the usual arguments for an abstract class apply as well in Ruby.  (At least, that's my interpretation of https://www.poodr.com/, which is my preferred reference on Ruby style.) 

But I'm not the one who has to live with this codebase.  I'll defer to the maintainers on this one.

@excpt, @anakinj, do you have opinions on @phlegx's suggestions?  I'm happy to implement it however you prefer.

anakinj:
I somewhat agree about the empty stubs. The naming is always hard, maybe in Ruby base classes are more common than abstracts.

But then again something that would bind these different types of keys together would be of benefit. Had a thought when this was discussed was that the list of [supported types ](https://github.com/jwt/ruby-jwt/blob/ac4f4e8c22ce0ca55e90390b640228c19bb3be36/lib/jwt/jwk.rb#L10) could be generated using something shared between these classes. Is that shared thing a base class or mixin i have no strong opinion on.

anakinj:
Added something small to the thing I'm currently working on in regards to the Base class for the different key types.
https://github.com/jwt/ruby-jwt/pull/375/commits/89247c4dbd1dccad536028e7648a0ab50f8283b9

And as @richardlarocque suggested I also removed the empty method stubs.

phlegx:
@richardlarocque can you please adapt your code to @anakinj base class and solve review issues?

richardlarocque:
I have reservations about rebasing on top of an uncommitted branch.

My goal in opening this PR was to get this functionality into the ruby-jwt upstream.  I built this PR so that it fit the style of the project as it existed at the time, so that it would be easy to merge.  My preference is that we merge this more or less as it is, in conformance with the old style.

If this project is in the middle of a refactor, and as a result we can't accept the PR as-is right now, that's OK.  I just had bad timing, I guess.  But I'd rather not get involved with the refactoring myself.  I'd prefer to wait until that refactoring lands in master and then rebase my change against that.

I don't want this feature's fate to be tied to that of some refactoring work.  I'd like to keep the two things separate.

I am willing to add support for the `include_private` flag on export and adding inheritance from `KeyAbstract`, though.  That stuff is in master, so I'd be OK with making this PR conform to those standards, if that is sufficient to get this PR merged.

d11wtq:
Just a comment to say thank you so much for doing this. We need this functionality too and exploring the alternative JWT gems, I didn't like what I saw in some of the solutions. Would love to see this one get across the line. I'm happy to help in any way I can do too ❤️

phlegx:
@richardlarocque all other PR have been merged and are in master. Can you please adapt the code style and options to RSA and HMAC? Thank you so much.

phlegx:
@anakinj @excpt should I open a new PR with this code changes, so that we can continue?

anakinj:
@richardlarocque do you mind someone continue with this to get it into the next release or are you planning on continuing on this on your own?

Would be really great to get this into master.

richardlarocque:
Sorry I'm late to the follow-up!  I can try to rebase in the next few days, but if one of you wants to beat me to it, feel free.

phlegx:
Nice work @richardlarocque! Thx

anakinj:
Thanks @richardlarocque for the contribution and patience.

phlegx:
@anakinj is a release planned?

anakinj:
@phlegx I would like to get the version in a shape that it would work with Ruby 2.1 and 2.2 as the .gemspec claims. Then the version could be the last version on the 2.x branch.

There is a few PRs open that addresses this issue.

@excpt hope you agree :)

PR #373, opened 2020-09-25
Support exporting RSA private keys

anakinj:
@phlegx @richardlarocque Got inspired by your work so made this thing that would allow the inclusion of the private key in the JWK.

### DIFF
--- a/lib/jwt/jwk/rsa.rb
+++ b/lib/jwt/jwk/rsa.rb
@@ -28,45 +28,68 @@ module JWT
         OpenSSL::Digest::SHA256.hexdigest(sequence.to_der)
       end
 
-      def export
-        {
+      def export(options = {})
+        ret = {
           kty: KTY,
           n: encode_open_ssl_bn(public_key.n),
           e: encode_open_ssl_bn(public_key.e),
           kid: kid
         }
+
+        return ret if options[:include_private] != true
+
+        ret.merge(
+          d: encode_open_ssl_bn(keypair.d),
+          p: encode_open_ssl_bn(keypair.p),
+          q: encode_open_ssl_bn(keypair.q),
+          dp: encode_open_ssl_bn(keypair.dmp1),
+          dq: encode_open_ssl_bn(keypair.dmq1),
+          qi: encode_open_ssl_bn(keypair.iqmp)
+        )
       end
 
       def encode_open_ssl_bn(key_part)
         ::Base64.urlsafe_encode64(key_part.to_s(BINARY), padding: false)
       end
 
-      def self.import(jwk_data)
-        jwk_n = jwk_data[:n] || jwk_data['n']
-        jwk_e = jwk_data[:e] || jwk_data['e']
-
-        raise JWT::JWKError, 'Key format is invalid for RSA' unless jwk_n && jwk_e
-
-        self.new(rsa_pkey(jwk_n, jwk_e))
-      end
-
-      def self.rsa_pkey(jwk_n, jwk_e)
-        key = OpenSSL::PKey::RSA.new
-        key_n = decode_open_ssl_bn(jwk_n)
-        key_e = decode_open_ssl_bn(jwk_e)
-
-        if key.respond_to?(:set_key)
-          key.set_key(key_n, key_e, nil)
-        else
-          key.n = key_n
-          key.e = key_e
+      class << self
+        def import(jwk_data)
+          self.new(rsa_pkey(*jwk_attrs(jwk_data, :n, :e, :d, :p, :q, :dp, :dq, :qi)))
         end
 
-        key
-      end
+        def jwk_attrs(jwk_data, *attrs)
+          attrs.map do |attr|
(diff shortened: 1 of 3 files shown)


anakinj:
Will look at those robot comments later. The amount of parameters is a little problematic for these complexity metrics

lib/jwt/jwk/rsa.rb
```diff
@@ -28,45 +28,68 @@ def kid
...
-        {
+      def export(options = {})
+        ret = {
           kty: KTY,
           n: encode_open_ssl_bn(public_key.n),
           e: encode_open_ssl_bn(public_key.e),
           kid: kid
         }
+
+        return ret if options[:include_private] != true
```
phlegx, line 39: Also check if the keypair has a private key.

`return ret unless private? && options[:include_private] == true`

phlegx (changes requested):
Should I also apply this changes (option include_private) to HMAC JWK?

lib/jwt/jwk/rsa.rb
anakinj: Good point. Will get back to this when I have the time.

anakinj:
> Should I also apply this changes (option include_private) to HMAC JWK?

If we can agree on the public export method API to be like this it could be already implemented for the new JWKs.

I went with the options hash to be able to maybe add/override something on the export easier in the future, dunno if that was necessary.

phlegx:
Options with a hash for export is the right way! I agree.

lib/jwt/jwk/rsa.rb
```diff
@@ -28,45 +28,68 @@ def kid
...
+          d: encode_open_ssl_bn(keypair.d),
+          p: encode_open_ssl_bn(keypair.p),
+          q: encode_open_ssl_bn(keypair.q),
+          dp: encode_open_ssl_bn(keypair.dmp1),
+          dq: encode_open_ssl_bn(keypair.dmq1),
+          qi: encode_open_ssl_bn(keypair.iqmp)
+        )
       end
 
       def encode_open_ssl_bn(key_part)
```
phlegx, line 51: Make this method private. So, we can expose the same set of methods in all other JWKs.

phlegx:
I propose to expose only a defined set of methods for all JWKs. All other methods should be private. An abstract class could help.

```ruby
module JWT
  module JWK
    class [NAME]
      attr_reader :keypair
      attr_reader :kid

      def initialize(keypair, kid = nil)
        # ...
      end

      def private?
        # ...
      end

      def public_key
        # ...
      end

      def export(options = {})
        # ...
      end

      class << self
        def import(jwk_data)
        end
      end
    end
  end
end

**Comment by phlegx** [comment]
@anakinj please review my code on PR #372. I have added a factory class that can be used by RSA and EC JWKs.

File: spec/jwk/rsa_spec.rb [hunk 3]
```diff
@@ -34,7 +34,7 @@
       it 'returns a hash with the public parts of the key' do
         expect(subject).to be_a Hash
         expect(subject).to include(:kty, :n, :e, :kid)
-        expect(subject).not_to include(:d)
+        expect(subject).not_to include(:d, :p, :dp, :dq,:qi)
```
richardlarocque, line 37: nit: whitespace between `:dq` and `:qi`.

spec/jwk/rsa_spec.rb
```diff
@@ -43,7 +43,7 @@
...
```
richardlarocque, line 46: nit: whitespace again.

spec/jwk/rsa_spec.rb
```diff
@@ -53,6 +53,15 @@
         expect { subject }.to raise_error(ArgumentError, 'keypair must be of type OpenSSL::PKey::RSA')
       end
     end
+
+    context 'when private key is requested' do
+      subject { described_class.new(keypair).export(include_private: true) }
+      let(:keypair) { rsa_key }
+      it 'returns a hash with the public AND private parts of the key' do
+        expect(subject).to be_a Hash
+        expect(subject).to include(:kty, :n, :e, :kid, :d, :p, :q, :dp, :dq,:qi)
```
richardlarocque, line 62: nit: also whitespace.

richardlarocque (commented):
This makes sense to me.

I don't pretend to understand the OpenSSL calls, but they seem fine.

I think there's an API question about what you want to do if someone passes `export_private` on a public key JWK.  You can raise an error or silently return the public key.  My preference is to raise an error, but you can argue it either way.

anakinj:
Thanks for the comments @richardlarocque. I hope to get back to this in the end of this week.

phlegx:
@anakinj @richardlarocque @excpt I suggest to first merge PR #372 that includes the abstract class. Then JWK RSA and EC can merge and adapt own PR to the abstract structure.

excpt:
#372 merged.

@anakinj I'd recommend a rebase for this pr.

anakinj:
Replaced by #375

PR #376, opened 2020-10-02
Disable Rails cops

anakinj:
Having the Rails Cops enabled gives false warnings about Rails specific syntax. For example it suggest using `delegate` even if the gem does not use [activesupport](https://guides.rubyonrails.org/active_support_core_extensions.html#delegate) that provides this functionality.


excpt (approved):

excpt:
🥳 Thabks again!

PR #381, opened 2020-10-23
Replace Travis CI with GitHub Actions (also favor openssl/rbnacl combinations over rails compatibility tests)

anakinj:
So far the CI has been checking the compatibility with different Rails versions but the risk of conflicts are slim and i'm arguing the value of the tests are minimal.

On other hand there has been some real issues with compatibility with openssl (#333), CI has always installed the latest openssl 2.x gem version. This is a little in conflict what the gem claims it's dependencies are, there is no mention that the gem requires a 2.x version of the openssl gem.

Im suggesting to drop all the tests that are involving Rails and focus on the things this gem actually uses, the openssl and signing features.

This PR will be a trial-and-error thing to test the travis configuration.


anakinj:
Also this PR allows the truffleruby build to fail. It seems to be a little flaky and giving false positives. Some discussion in #368 related to that.

anakinj:
And we are adding Ruby 2.7 + ruby-head builds

anakinj:
@excpt What do you think about this? Think it would be valuable to ensure the compatibility with the rubies the gem support. The support for older EOL rubies could then be dropped in the next bigger release. Maybe 2.3.0?

anakinj:
TravisCI is really giving OSS projects a hard time. The build [queue](https://www.traviscistatus.com/#system-metrics) is hours long in the week. 

So in all this CI tuning I converted the original change into a GitHub actions workflow, the tests just take a few minutes to execute. Also the tests can be executed in your own fork before creating a PR. 

The original statement is still valid (Favor different cryptolib combinations over Rails compatibility).

An example on how the CI result would look like can be found from [here](https://github.com/anakinj/ruby-jwt/actions/runs/396946566) 

Now this also reveals that Ruby 2.3 tests without the separate openssl gem are failing because of #386

excpt (approved):

excpt:
@anakinj Thank you very much for the work to convert the CI to github Actions.

Dropping ruby support for older versions should not be a problem. 
I think we should keep an eye on the official supported list. 

https://www.ruby-lang.org/en/downloads/branches/

Dropping ruby versions older than 2.4 shouldn't be a problem but this discussion should continue in a new issue.

Issue #361, opened 2020-07-11
ArgumentError: invalid base64 while calling JWT::JWK.import(hash)


AbhishekJadav2:
My current ruby version is **2.2.2** which is greater than minimal ruby version requirement (>=2.1) as per .gemspec file **spec.required_ruby_version = '>= 2.1'*

Here's the full Stack Trace.
```
ArgumentError: invalid base64
	from /home/viraj/.rvm/rubies/ruby-2.2.2/lib/ruby/2.2.0/base64.rb:77:in `unpack'
	from /home/viraj/.rvm/rubies/ruby-2.2.2/lib/ruby/2.2.0/base64.rb:77:in `strict_decode64'
	from /home/viraj/.rvm/rubies/ruby-2.2.2/lib/ruby/2.2.0/base64.rb:95:in `urlsafe_decode64'
	from /home/viraj/.rvm/gems/ruby-2.2.2/gems/jwt-2.2.1/lib/jwt/jwk/rsa.rb:40:in `import'
	from /home/viraj/.rvm/gems/ruby-2.2.2/gems/jwt-2.2.1/lib/jwt/jwk.rb:19:in `import'
	from (irb):21
	from /home/viraj/.rvm/gems/ruby-2.2.2/gems/railties-4.2.3/lib/rails/commands/console.rb:110:in `start'
	from /home/viraj/.rvm/gems/ruby-2.2.2/gems/railties-4.2.3/lib/rails/commands/console.rb:9:in `start'
	from /home/viraj/.rvm/gems/ruby-2.2.2/gems/railties-4.2.3/lib/rails/commands/commands_tasks.rb:68:in `console'
	from /home/viraj/.rvm/gems/ruby-2.2.2/gems/railties-4.2.3/lib/rails/commands/commands_tasks.rb:39:in `run_command!'
	from /home/viraj/.rvm/gems/ruby-2.2.2/gems/railties-4.2.3/lib/rails/commands.rb:17:in `<top (required)>'
	from /home/viraj/.rvm/gems/ruby-2.2.2/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:274:in `require'
	from /home/viraj/.rvm/gems/ruby-2.2.2/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:274:in `block in require'
	from /home/viraj/.rvm/gems/ruby-2.2.2/gems/activesupport-4.2.3/lib/active_support/dependencies.rb:240:in `load_dependency'
```

[...]

PR #382, opened 2020-10-25
Support JWKs for pre 2.3 rubies

anakinj:
This should fix the issue in #361. Even if the Rubies are EOL and stone-age we should probably still support them while the gemspec claims to support `>= 2.1`

Part of the problem is that the CI has not been executed for pre 2.3 versions in a while, was thinking we could maybe add the older rubies to the test matrix in #381 before we remove the support completely in the next major version of this gem.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,10 +24,12 @@ jobs:
           - gemfiles/rbnacl.gemfile
         experimental: [false]
         include:
-#          - ruby: 2.1
-#            experimental: true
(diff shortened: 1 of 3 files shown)


anakinj:
2x Truffleruby builds failed because of something random. The flaky tests runs are also tagged as `allow_failures` in #381

anakinj:
Merged this to have working CI for 2.1 and 2.2 while working on some refactorings on the JWK parts of the gem

PR #385, opened 2020-12-01
Define the secp256r1 curve

anakinj:
This should fix #367 

Would be super to try this out @m-zielinski or @obulkin. My OpenSSL version was not capable of reproducing the issue.
`OpenSSL::PKey::EC.new('secp256r1')` threw a `OpenSSL::PKey::ECError (invalid curve name)`


obulkin:
Thanks, @anakinj. Been rather busy lately, but I'll try to validate this after the holidays.

In our case, the problematic key was pre-generated using `openssl ecparam -name prime256v1`. You could try to reproduce the issue this way instead, although it may still come down to have the right OpenSSL version.

anakinj:
Im gonna merge this. The change is just adding support for that old name.

PR #386, opened 2020-12-02
Make JWT::JWK::EC compatible with Ruby 2.3

anakinj:
Seems that the `OpenSSL::PKey::EC#generate_key!` method is not available in Ruby 2.3 without installing the openssl gem.

This other PR that is making the CIactually test the supported Rubies and combinations revealed that:
https://travis-ci.org/github/jwt/ruby-jwt/jobs/747060754

### DIFF
--- a/lib/jwt/jwk/ec.rb
+++ b/lib/jwt/jwk/ec.rb
@@ -4,7 +4,7 @@ module JWT
   module JWK
     class EC < KeyBase
       extend Forwardable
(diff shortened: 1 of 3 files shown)


anakinj:
@richardlarocque Are you OK with these changes. Unfortunately think we need to support the EOL rubies as long as the gemspec claims we do. Hoping to drop support for the oldies soon

anakinj:
I'll go on and merge this to get the tests green.

PR #388, opened 2020-12-04
JWK OKP support

anakinj:
Started to look into supporting the Ed25519 algorithm for the JWKs (https://tools.ietf.org/html/rfc8037)

Also while doing this I realized that not all keys come in pairs so therefore was thinking to deprecate the following methods

```
::JWT::JWK::RSA#keypair
::JWT::JWK::RSA#private_key
::JWT::JWK::RSA#public_key
```

To favor the more specific methods
```
::JWT::JWK::[RSA/EC/HMAC/OKP]#verify_key
::JWT::JWK::[RSA/EC/HMAC/OKP]#signing_key
```

This also affects the `::JWT::JWK::EC` and `::JWT::JWK::HMAC` APIs. But since these have not been in any official release we could maybe just not support the old methods.

I'll work on this whenever I have the time and energy. Comments are welcome whenever!

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,1 +1,2 @@
 --color
+--require spec_helper
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -59,7 +59,7 @@ Metrics/AbcSize:
   Max: 20
 
 Metrics/ClassLength:
-  Max: 100
+  Max: 110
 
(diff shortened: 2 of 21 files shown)


anakinj:
Need to redo the JWK classes apis to make any sense of it. Stay tunded for some refactoring...

PR #390, opened 2020-12-14
Enable rubocop locally and on ci

anakinj:
Been struggling with sourcelevel-bot spamming PRs with comments about linting issues like extra whitespaces and such.

This PR adds the possibility to run rubocop locally to catch issues early in the process.

The version should be the same as Codeclimate is using(https://docs.codeclimate.com/docs/rubocop)


anakinj:
@excpt Im thinking of merging this to improve the local experience when developing the gem.

PR #393, opened 2020-12-21
Fix Time.now casting causing false positive in verify_at

### LINKED ISSUE #353: Issued At RFC type mismatch causing false postives on verify_at
Hello there,

We had some troubles with `ruby-jwt` in production recently. The reason was that we put the unix timestamp as seconds since epoch in `iat`, but inside `verify_iat` ruby-jwt casts it using `.to_f`.

When comparing `iat` against `Time.now.to_f` you compare an int (rounded to `.0`) to a full decimal version, making `iat.to_f > Time.now.to_f` trigger a false positive.

The specification in https://tools.ietf.org/html/rfc7519#section-4.1.6 states that the `iat` should be a `NumericDate` as defined in RFC7519:

> **NumericDate**
>       A JSON numeric value representing the number of seconds from
>       1970-01-01T00:00:00Z UTC until the specified UTC date/time,
>       ignoring leap seconds.  This is equivalent to the IEEE Std 1003.1,
>       2013 Edition [POSIX.1] definition "**Seconds Since the Epoch**", in
>       which each day is accounted for by exactly 86400 seconds, other
>       than that non-integer values can be represented.  See RFC 3339
>       [RFC3339] for details regarding date/times in general and UTC in
>       particular.

The fix for this should be trivial (replacing `to_f` with `to_i`) but before submitting a pull request I'd like to check with you if that fix makes sense.

kevinreedy:
Fixes #353

### DIFF
--- a/lib/jwt/verify.rb
+++ b/lib/jwt/verify.rb
@@ -45,7 +45,7 @@ module JWT
       return unless @payload.include?('iat')
 
       iat = @payload['iat']
-      raise(JWT::InvalidIatError, 'Invalid iat') if !iat.is_a?(Numeric) || iat.to_f > Time.now.to_f
+      raise(JWT::InvalidIatError, 'Invalid iat') if !iat.is_a?(Numeric) || iat.to_i > Time.now.to_i
     end
 
     def verify_iss


anakinj:
Im wondering if the condition should be inclusive (`>=`) to handle cases where the verification happen on the same second the token has ben issued.

Also a test verifying the "same-second" token would be good. Mocking the `Time.now.to_i` in a test could be a way to tackle the different cases.

anakinj:
Structure has changed quite a lot since this was touched the last time. Closing this one for now. Please reopen/refresh if there is a need.

PR #402, opened 2021-01-25
documenting calling verify_jti callback with 2 arguments in the readme

HoneyryderChuck:
Had to dive into the source code to find this one. Maybe the next one will not have to do the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,6 +412,8 @@ begin
   #decoded_token = JWT.decode token, hmac_secret, true, { verify_jti: true, algorithm: 'HS256' }
   # Alternatively, pass a proc with your own code to check if the JTI has already been used
   decoded_token = JWT.decode token, hmac_secret, true, { verify_jti: proc { |jti| my_validation_method(jti) }, algorithm: 'HS256' }
+  # or
+  decoded_token = JWT.decode token, hmac_secret, true, { verify_jti: proc { |jti, payload| my_validation_method(jti, payload) }, algorithm: 'HS256' }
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -412,6 +412,8 @@ begin
   #decoded_token = JWT.decode token, hmac_secret, true, { verify_jti: true, algorithm: 'HS256' }
   # Alternatively, pass a proc with your own code to check if the JTI has already been used
   decoded_token = JWT.decode token, hmac_secret, true, { verify_jti: proc { |jti| my_validation_method(jti) }, algorithm: 'HS256' }
+  # or
+  decoded_token = JWT.decode token, hmac_secret, true, { verify_jti: proc { |jti, payload| my_validation_method(jti, payload) }, algorithm: 'HS256' }
```
anakinj, line 416: Maybe we could just replace the existing example with this one? I mean if someone does not need the `payload` it can just be omitted.

HoneyryderChuck: Given that the call supports both signatures, I think both should be documented. There's probably enough people using and needing only the first approach, I'd say.

anakinj: I am convinced this is good lite this. 

Think you need to rebase so I can merge this. There were another MR related to the README merged just a moment ago.

anakinj (approved):

anakinj:
Thanks @HoneyryderChuck. I was able to merge this as-is.

HoneyryderChuck:
Thank you @anakinj !!

PR #404, opened 2021-02-04
Document function and add tests for verify claims method

yasonk:
Had to dig through the code to find out how to disable verifications. Decided to add tests for the `verify_claims` method because it is the one using the options to determine which claims to verify.


anakinj (approved):

anakinj:
Looks great. Thanks @yasonk

Issue #334, opened 2019-09-20
Wrong `alg` used in header for Ed25519 keys


zblach:
The example tokens in this library use `"alg": "ED25519"`, but the related spec seems to suggest that it be `"EdDSA"` instead.

https://tools.ietf.org/html/rfc8037#appendix-A.4

anakinj:
I think you are correct. Now I wonder how we could make this right. The feature has been out in the wild for a pretty long time already and encoded tokens get this incorrect `alg` header. 

I would suggest:
- Not to change the generated `alg` header (The decoding should not care about the given header anyway)
- Start supporting the correct algorithm (EdDSA) in the decode method

mattgibson:
>Not to change the generated alg header (The decoding should not care about the given header anyway)

I just ran into this when trying to use a Ruby-generated token in an Elixir app. The Elixir library is using the `alg` header to choose the decryption algorithm and is case-sensitive, so `ED25519` causes a signature error as it can't work out how to handle it, but `Ed25519` works.

anakinj:
Just out of curiosity, what Elixir library is this?

My biggest concern is that if it is just changed to something else for example old ruby-jwt<-> new ruby-jwt compatibility will at least break.

anakinj:
If you are in control of the token generation you could change the constant controlling this locally:
`JWT::Algos::Eddsa::SUPPORTED = ["Ed25519"]`

mattgibson:
I was using [Joken](https://github.com/joken-elixir/joken), but the underlying low level library it uses is [JOSE](https://github.com/potatosalad/erlang-jose)

mattgibson:
>If you are in control of the token generation you could change the constant controlling this locally:
JWT::Algos::Eddsa::SUPPORTED = ["Ed25519"]

Yes, I found that and it worked when generating the tokens. I think for now I'm going to fork the gem and ensure it will decrypt both `Ed25519` and `ED25519` tokens. Thanks for the pointer!

anakinj:
But apparently this Joken component is expecting the EdDSA algorithms to have the exact curve used in the `alg` field. So changing the alg to `EdDSA` will not make this gem compatible either.

mattgibson:
Ah, that's a good point. Possibly Joken is not implementing the RFC correctly either.

anakinj:
Btw. I think the easiest workaround is just to monkeypatch the `JWT::Algos::Eddsa::SUPPORTED` constant. No need for forks.

```
[...]

PR #405, opened 2021-02-04
Ignore casing of algorithm

johnnyshields:
(Updated 2/6/2021)

I am have to connect to a third party source for JWTs which transmits the algorithm as "Ed25519". This does not match the expected value in Ruby "ED25519". To remedy this:

- Ignore casing of algorithm on input
- Output correct casing of algorithm when encoding
- Consider "Ed25519" as the correct casing, as per [RFC#8037](https://tools.ietf.org/html/rfc8037) as well as the [Go](https://github.com/gbrlsnchs/jwt/blob/f423236368671cbb9216d3893ff14094d29407c6/verify_test.go#L55), [Erlang](https://github.com/potatosalad/erlang-jose/blob/c7aad59f78e16659ea9aaa8953303275c4949eef/src/jwk/jose_jwk_kty_okp_ed25519.erl#L49), [Haskell](https://github.com/tekul/jose-jwt/blob/718fe38326a59e7310a49f166a8ad36f3377727d/tests/jwks.json#L11), etc. JWT implementations.

In order to achieve this in a clean manner, I've extracted out a `JWT::Algos` module which uses a more efficient lookup (pre-indexed) to find the algo. I've also introduced `JWT::Algos::None` which is distinct from `JWT::Algos::Unsupported`

### DIFF
--- a/lib/jwt/algos.rb
+++ b/lib/jwt/algos.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'jwt/algos/hmac'
+require 'jwt/algos/eddsa'
+require 'jwt/algos/ecdsa'
+require 'jwt/algos/rsa'
+require 'jwt/algos/ps'
+require 'jwt/algos/unsupported'
+
+# JWT::Signature module
+module JWT
+  # Signature logic for JWT
+  module Algos
+    extend self
+    ALGOS = [
+      Algos::Hmac,
+      Algos::Ecdsa,
+      Algos::Rsa,
+      Algos::Eddsa,
+      Algos::Ps,
+      Algos::Unsupported
+    ].freeze
+
+    def find(algorithm)
+      ALGOS.each do |alg|
+        code = alg.const_get(:SUPPORTED).find {|a| a.upcase == algorithm.upcase }
+        return [alg, code] if code
+      end
+      nil
+    end
+  end
+end
--- a/lib/jwt/algos.rb
+++ b/lib/jwt/algos.rb
@@ -13,6 +13,7 @@ module JWT
   # Signature logic for JWT
   module Algos
     extend self
+
     ALGOS = [
       Algos::Hmac,
       Algos::Ecdsa,
--- a/lib/jwt/algos.rb
+++ b/lib/jwt/algos.rb
@@ -5,6 +5,7 @@ require 'jwt/algos/eddsa'
 require 'jwt/algos/ecdsa'
 require 'jwt/algos/rsa'
 require 'jwt/algos/ps'
+require 'jwt/algos/none'
 require 'jwt/algos/unsupported'
 
 # JWT::Signature module
@@ -18,15 +19,26 @@ module JWT
       Algos::Rsa,
       Algos::Eddsa,
       Algos::Ps,
+      Algos::None,
       Algos::Unsupported
     ].freeze
 
     def find(algorithm)
-      ALGOS.each do |alg|
-        code = alg.const_get(:SUPPORTED).find {|a| a.upcase == algorithm.upcase }
-        return [alg, code] if code
+      indexed[algorithm && algorithm.downcase]
+    end
+
+    private
+
+    def indexed
+      @indexed ||= begin
+        algos = ALGOS.dup
+        fallback = [algos.pop, nil]
+        algos.each_with_object(Hash.new(fallback)) do |alg, hash|
+          alg.const_get(:SUPPORTED).each do |code|
+            hash[code.downcase] = [alg, code]
+          end
+        end
       end
-      nil
     end
   end
 end
--- a/lib/jwt/algos.rb
+++ b/lib/jwt/algos.rb
@@ -32,9 +32,8 @@ module JWT
 
     def indexed
       @indexed ||= begin
-        algos = ALGOS.dup
-        fallback = [algos.pop, nil]
-        algos.each_with_object(Hash.new(fallback)) do |alg, hash|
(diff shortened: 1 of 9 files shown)


anakinj:
There might be something wrong with how the gem handles the Ed25519 cases (https://github.com/jwt/ruby-jwt/issues/334).

In general i think that making the algorithms case insensitive is maybe not a good idea. The RFCs clearly states what the values are expected to be (https://tools.ietf.org/html/rfc7518#page-9).

johnnyshields:
> The RFCs clearly states what the values are expected to be (https://tools.ietf.org/html/rfc7518#page-9).

The RFC does not say anything about case sensitivity. It's reasonable to expect that there will case insensitive JWT implementations in the wild. Perhaps we can always encode to correct casing (uppercase), but decode irrespective of case?

johnnyshields:
@anakinj I've raised PR #406 which is specific to "Ed25519". That PR includes examples of how other libs are using "Ed25519" (lowercase "d"). Please kindly merge either this PR or #406 .

anakinj:
I agree that that the parameter given in the `decode` method could be case insensitive. 

Just to compare the OpenSSL API is case insensitive and allows both uppercase and lowercase:
```
OpenSSL::Digest.new('SHA512') == OpenSSL::Digest.new('sha512')
=> true
```

anakinj:
Im fine with either just doing the change in #406 or then make some adjustments to this that the `alg` value in the token do not change but the encode and decode methods understands the algorithm in any given case

johnnyshields:
OK, I've updated this PR to show what it would look like to add casing coercion (e.g. "hs256" becomes "HS256" when encoding). There are still some failing specs related to unsupported case. If you like this implementation I will fix the failing specs over the weekend.

lib/jwt/algos.rb
```diff
@@ -0,0 +1,32 @@
...
+require 'jwt/algos/ecdsa'
+require 'jwt/algos/rsa'
+require 'jwt/algos/ps'
+require 'jwt/algos/unsupported'
+
+# JWT::Signature module
+module JWT
+  # Signature logic for JWT
+  module Algos
+    extend self
```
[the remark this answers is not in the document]
johnnyshields: Doing this change breaks the behavior

lib/jwt/algos.rb
```diff
@@ -0,0 +1,45 @@
...
+    def find(algorithm)
+      indexed[algorithm && algorithm.downcase]
+    end
+
+    private
+
+    def indexed
+      @indexed ||= begin
+        algos = ALGOS.dup
+        fallback = [algos.pop, nil]
```
anakinj, line 36: Now the `Algos::Unsupported` is expected to be the last entry in the array. What is the fallback would just be hardcoded? Also then there would be no need for duplicating the array.

Or was something using that `ALGOS` constant directly still?

lib/jwt/algos/eddsa.rb
```diff
@@ -3,7 +3,7 @@ module Algos
     module Eddsa
       module_function
 
-      SUPPORTED = %w[ED25519].freeze
+      SUPPORTED = %w[Ed25519].freeze
```
anakinj, line 6: Does this change the value that is added to the `alg` header in the payload?

anakinj (changes requested):
Looks really great. I added a few questions

spec/jwt_spec.rb
```diff
@@ -510,4 +510,25 @@
...
   end
+
+  context 'algorithm case insensitivity' do
+    let(:payload) { { 'a' => 1, 'b' => 'b' } }
+
+    it 'ignores algorithm casing during encode/decode' do
+      enc = JWT.encode(payload, '', 'hs256')
+      expect(JWT.decode(enc, '')).to eq([payload, { 'alg' => 'HS256'}])
+
+      enc = JWT.encode(payload, data[:rsa_private], 'ps384')
```
anakinj, line 521: The PS algorithms require OpenSSL 2.1. So until we drop support for older rubies the tests that have OpenSSL2.1 dependencies need to be separated.

lib/jwt/algos.rb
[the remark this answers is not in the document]
anakinj: I myself prefer using `class << self` in these cases.

lib/jwt/algos/eddsa.rb
anakinj: Not 100% sure but I think that the alg header when using this algorithm + curve needs to be `EdDSA` as #334 suggests.

I suggest fixing this as a separate thing. Also if there are some other services/components that are assigning the alg header as `Ed25519` it would be nice to know the reasoning behind that.

johnnyshields: @anakinj if so, that should be a separate PR.

anakinj: Yes, in a separate PR and maybe it needs to go into the next major release.

Changing the generated `alg` is a breaking change and we should be a little careful, even if the current way would be totally wrong there could be some Ruby JWT-JWT implementations trusting that the alg is what it is.

johnnyshields: Are you ok with the change above?

anakinj: I'd rather not change the casing on the algorithm, but fix it properly in the next major release.

johnnyshields: Ok, reverted.

anakinj (approved):

anakinj:
Thank you for your contribution.

PR #406, opened 2021-02-05
Support alternate casing "Ed25519" (lowercase "d")

johnnyshields:
Currently the system requires "ED25519". However, [RFC#8037](https://tools.ietf.org/html/rfc8037) refers to it clearly as "Ed25519" (lowercase "d"), and this casing is used in various other JWT libs such as [Go](https://github.com/gbrlsnchs/jwt/blob/f423236368671cbb9216d3893ff14094d29407c6/verify_test.go#L55), [Erlang](https://github.com/potatosalad/erlang-jose/blob/c7aad59f78e16659ea9aaa8953303275c4949eef/src/jwk/jose_jwk_kty_okp_ed25519.erl#L49), and [Haskell](https://github.com/tekul/jose-jwt/blob/718fe38326a59e7310a49f166a8ad36f3377727d/tests/jwks.json#L11).

### DIFF
--- a/lib/jwt/algos/eddsa.rb
+++ b/lib/jwt/algos/eddsa.rb
@@ -3,7 +3,7 @@ module JWT
     module Eddsa
       module_function
 
-      SUPPORTED = %w[ED25519].freeze
+      SUPPORTED = %w[Ed25519 ED25519].freeze
 
       def sign(to_sign)
         algorithm, msg, key = to_sign.values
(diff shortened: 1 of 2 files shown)


johnnyshields:
@anakinj please take a look

anakinj:
This looks great. Think that the tests are failing because they are expecting `Ed25519_private` and `Ed25519_private` to exist in the hash on https://github.com/jwt/ruby-jwt/blob/6ff11dcb89aee49857ccf6e876b2b386cdbab895/spec/jwt_spec.rb#L40

johnnyshields:
Specs fixed

anakinj:
I'll close this one. #405 gives the case insensitivity and we need a proper fix for #334

Issue #410, opened 2021-03-31
NoMethodError: undefined method `zero?' for nil:NilClass if JWT has no 'alg' field


bouk:
For example:

```ruby
 JWT.decode('e30K.e30K.e30K')
=> NoMethodError: undefined method `zero?' for nil:NilClass
```

I'd expect this to throw a `JWT::IncorrectAlgorithm`, but casecmp here: https://github.com/jwt/ruby-jwt/blob/2cea14fdae439773fafc59640178e5cf7a0af8a4/lib/jwt/decode.rb#L46 returns `nil` if alg is not set, so we get an unexpected NoMethodError.

PR #411, opened 2021-03-31
Raise IncorrectAlgorithm if token has no alg header

bouk:
Closes https://github.com/jwt/ruby-jwt/issues/410


anakinj:
Thanks for the fix.

As long as the alg header is required by the validation i think this is a great change. In general I think the alg header should be of almost no relevance and maybe we should make it less important when validating the signature.

anakinj (approved):

PR #416, opened 2021-04-20
Fix references for v2.2.3 on CHANGELOG

vyper:
Some references were pointing to 2.2.3 instead of v2.2.3.


anakinj (approved):

anakinj:
Thanks for the fix

PR #417, opened 2021-04-22
Fixed a typo about class name 

mai-f:
Fixed from Jwt::JWKError to JWT::JWKError.


anakinj (approved):

anakinj:
Thanks for the fix. Adding a few teststo test these error cases would be of value.

PR #418, opened 2021-05-06
Add metadata for RubyGems

nickhammond:


ruby-jwt.gemspec
```diff
@@ -14,6 +14,10 @@ Gem::Specification.new do |spec|
   spec.homepage = 'https://github.com/jwt/ruby-jwt'
   spec.license = 'MIT'
   spec.required_ruby_version = '>= 2.1'
+  spec.metadata = {
+    'bug_tracker_uri' => 'https://github.com/jwt/ruby-jwt/issues',
+    'changelog_uri'   => 'https://github.com/jwt/ruby-jwt/blob/master/CHANGELOG.md'
```
anakinj, line 19: Would it make sense to pin the changelog to the released version?

```suggestion
    'changelog_uri'   => "https://github.com/jwt/ruby-jwt/blob/v#{JWT.gem_version}/CHANGELOG.md"
```

anakinj:
Thanks for the improved metadata content. Added one comment, dunno if it makes sense to be that specific...

ruby-jwt.gemspec
nickhammond: I've seen both ways and I do like how it's easier to go directly to the changes for that version, I'd say that's your call though.

anakinj: If you're fine with going with the version specific link then let's do that.

anakinj (approved):

anakinj:
Thank you for the contribution. Highly appreciated 👍

PR #419, opened 2021-05-12
Initialize jwks with an empty hash if jwk_loader is given

hobbypunk90:


anakinj:
What is the reasoning behind this change? 

Think having an empty hash as the default value of the instance variable will make the loader to never trigger because of this:

https://github.com/jwt/ruby-jwt/blob/master/lib/jwt/jwk/key_finder.rb#L39

hobbypunk90:
i see the problem 🙈  our problem is, without an initilization, the keys are never revolved. It raises an exception at
https://github.com/jwt/ruby-jwt/blob/master/lib/jwt/jwk/key_finder.rb#L50
but its better to catch that error and return an empty array 👍

anakinj:
Are you giving a keyloader or something else as the jwks parameter?

anakinj:
Think that this class has some problems if something else than a `Hash` or an object responding to `:call` is given. In those cases it has no chances of loading anything into the `@jwks` instance variable and could as well be initialized as an empty hash.

anakinj:
Maybe we could have an empty hash as the default if the loader callback returns nil, something like this?

```
def load_keys(opts = {})
  @jwks = @jwk_loader.call(opts) || {}
end
```

hobbypunk90:
@anakinj sorry 🤦  we found the problem, it was in our loader proc.
we only loaded keys if the option invalidate was set.

anakinj:
Great that you found the issue.

PR #423, opened 2021-06-04
Tests for iat verification behaviour

anakinj:
Wrote some test related to how the iat claim is verified by the gem at the moment.


cabello:
My clock was slightly out of sync and I was getting `iat` validation errors, I am disabling it for now, but that caught me off guard and it was fixed once I re-synced my computer clock, I worry about our production systems getting out of sync like that and causing a major auth incident.

PR #425, opened 2021-06-17
Add Support to be able to verify from multiple keys

ritikesh:
**Problem -** 

During a security incident, if one has to rotate secrets shared with 3rd party apps, one has to ensure backward compatibility where both the old and new secrets continue to work.

**Solution -** 
Have made changes to iterate over an array of keys while verifying signature. This can be used by returning an array of keys from the `find_key` block.

PS: Will add specs if changes are ok.

### DIFF
--- a/lib/jwt/decode.rb
+++ b/lib/jwt/decode.rb
@@ -23,6 +23,7 @@ module JWT
       validate_segment_count!
       if @verify
         decode_crypto
+        set_key
         verify_signature
         verify_claims
       end
@@ -33,18 +34,32 @@ module JWT
     private
 
     def verify_signature
+      return unless @key || @verify
+
+      return if none_algorithm?
+
+      raise JWT::DecodeError, 'No verification key available' unless @key
+
+      return if Array(@key).any? { |key| verify_signature_for?(key) }
+
+      raise(JWT::VerificationError, 'Signature verification failed')
+    end
+
+    def set_key
       raise(JWT::IncorrectAlgorithm, 'An algorithm must be specified') if allowed_algorithms.empty?
-      raise(JWT::IncorrectAlgorithm, 'Token is missing alg header') unless header['alg']
+      raise(JWT::IncorrectAlgorithm, 'Token is missing alg header') unless algorithm
       raise(JWT::IncorrectAlgorithm, 'Expected a different algorithm') unless options_includes_algo_in_header?
 
       @key = find_key(&@keyfinder) if @keyfinder
       @key = ::JWT::JWK::KeyFinder.new(jwks: @options[:jwks]).key_for(header['kid']) if @options[:jwks]
+    end
 
-      Signature.verify(header['alg'], @key, signing_input, @signature)
+    def verify_signature_for?(key)
+      Signature.verify(algorithm, key, signing_input, @signature)
     end
 
     def options_includes_algo_in_header?
-      allowed_algorithms.any? { |alg| alg.casecmp(header['alg']).zero? }
+      allowed_algorithms.any? { |alg| alg.casecmp(algorithm).zero? }
     end
 
     def allowed_algorithms
@@ -65,8 +80,10 @@ module JWT
 
     def find_key(&keyfinder)
       key = (keyfinder.arity == 2 ? yield(header, payload) : yield(header))
-      raise JWT::DecodeError, 'No verification key available' unless key
-      key
+      # key can be of type [string, nil, OpenSSL::PKey, Array]
+      return key if key && !Array(key).empty?
+
+      raise JWT::DecodeError, 'No verification key available'
     end
 
     def verify_claims
@@ -77,7 +94,7 @@ module JWT
     def validate_segment_count!
       return if segment_length == 3
       return if !@verify && segment_length == 2 # If no verifying required, the signature is not needed
-      return if segment_length == 2 && header['alg'] == 'none'
+      return if segment_length == 2 && none_algorithm?
 
       raise(JWT::DecodeError, 'Not enough or too many segments')
     end
@@ -86,8 +103,16 @@ module JWT
       @segments.count
     end
 
+    def none_algorithm?
+      algorithm.casecmp('none').zero?
+    end
+
     def decode_crypto
       @signature = JWT::Base64.url_decode(@segments[2] || '')
+    end
+
+    def algorithm
+      header['alg']
     end
 
     def header
--- a/lib/jwt/signature.rb
+++ b/lib/jwt/signature.rb
@@ -23,13 +23,8 @@ module JWT
     end
 
     def verify(algorithm, key, signing_input, signature)
-      return true if algorithm.casecmp('none').zero?
-
-      raise JWT::DecodeError, 'No verification key available' unless key
-
       algo, code = Algos.find(algorithm)
-      verified = algo.verify(ToVerify.new(code, key, signing_input, signature))
-      raise(JWT::VerificationError, 'Signature verification raised') unless verified
+      algo.verify(ToVerify.new(code, key, signing_input, signature))
     rescue OpenSSL::PKey::PKeyError
       raise JWT::VerificationError, 'Signature verification raised'
     ensure
--- a/spec/jwt_spec.rb
+++ b/spec/jwt_spec.rb
@@ -357,6 +357,118 @@ RSpec.describe JWT do
   end
 
   context 'Verify' do
+    context 'when key given as an array with multiple possible keys' do
+      let(:payload) { { 'data' => 'data'} }
+      let(:token)   { JWT.encode(payload, secret, 'HS256') }
+      let(:secret)  { 'hmac_secret' }
+
+      it 'should be able to verify signature when block returns multiple keys' do
+        decoded_token = JWT.decode(token, nil, true, { algorithm: 'HS256' }) do
+          ['not_the_secret', secret]
+        end
+        expect(decoded_token.first).to eq(payload)
+      end
+
+      it 'should be able to verify signature when multiple keys given as a parameter' do
+        decoded_token = JWT.decode(token, ['not_the_secret', secret], true, { algorithm: 'HS256' })
+        expect(decoded_token.first).to eq(payload)
+      end
+
+      it 'should fail if only invalid keys are given' do
+        expect do
+          JWT.decode(token, ['not_the_secret', 'not_the_secret_2'], true, { algorithm: 'HS256' })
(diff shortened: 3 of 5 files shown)


lib/jwt/decode.rb
```diff
@@ -65,7 +71,7 @@ def allowed_algorithms
 
     def find_key(&keyfinder)
       key = (keyfinder.arity == 2 ? yield(header, payload) : yield(header))
-      raise JWT::DecodeError, 'No verification key available' unless key
+      raise JWT::DecodeError, 'No verification key available' unless key && key.empty?
```
anakinj, line 74: Think this is assuming the key returned from the block is an array. So if a single string is returned i think this will fail with a `NoMethodError`

anakinj (changes requested):

anakinj:
Thanks for the proposal, first impression is that I think this is a pretty good idea. Have to digest it a little still if there are potential problems with this kind of behaviour. A little finetuning to get the specs passing and maybe adding a few new tests.

lib/jwt/signature.rb
```diff
@@ -28,8 +28,7 @@ def verify(algorithm, key, signing_input, signature)
       raise JWT::DecodeError, 'No verification key available' unless key
 
       algo, code = Algos.find(algorithm)
-      verified = algo.verify(ToVerify.new(code, key, signing_input, signature))
-      raise(JWT::VerificationError, 'Signature verification raised') unless verified
+      algo.verify(ToVerify.new(code, key, signing_input, signature))
     rescue OpenSSL::PKey::PKeyError
       raise JWT::VerificationError, 'Signature verification raised'
```
anakinj, line 33: I wonder in what cases this error is raised. Could be that this feature will only work for symmetric algorithms

lib/jwt/decode.rb
ritikesh: I checked in ruby 2.4.7+. `String#empty?` and `Array#empty?` are both valid methods. Am I missing something obvious here? 🤔

lib/jwt/signature.rb
ritikesh: in my local testing, I could see that the error was being raised in cases where I passed an invalid secret. JWT defaults to `HS256`, so I might have assumed that it raises those exceptions by default when an invalid secret is being passed -

```ruby
str = JWT.encode({iss: 'me'}, 'passcode')
# with correct secret
JWT.decode(str) { |headers, payload| puts(headers); 'passcode' }
{"alg"=>"HS256"} # headers
=> [{"iss"=>"me"}, {"alg"=>"HS256"}]

# with incorrect secret
JWT.decode(str) { |headers, payload| puts(headers); 'passcoded' }
{"alg"=>"HS256"} # headers
=> Signature verification raised (JWT::VerificationError)

JWT.decode(str)
=> Signature verification raised (JWT::VerificationError)
```

ritikesh:
> Thanks for the proposal, first impression is that I think this is a pretty good idea. Have to digest it a little still if there are potential problems with this kind of behaviour. A little finetuning to get the specs passing and maybe adding a few new tests.

Thanks! I'll add some new tests tomorrow and fix the existing specs as well.

lib/jwt/decode.rb
anakinj: You're right. This works just fine.

lib/jwt/signature.rb
anakinj: Sorry a little misleading comment there, was referring to the `OpenSSL::PKey::PKeyError`. And that will probably be raised when an invalid public key is given to the verifier. For example for the RSA algorithms. But not covering those cases is probably just fine, we can iterate more if needed.

spec/jwk/ec_spec.rb
```diff
@@ -18,7 +18,7 @@
       let(:keypair) { ec_key.tap { |x| x.private_key = nil } }
       it 'creates an instance of the class' do
         expect(subject).to be_a described_class
-        expect(subject.private?).to eq false
+        expect(subject.private?).to eq true
```
anakinj, line 21: This is related to a change in the OpenSSL lib. Think the test is fixed in the current master branch.

spec/jwt_spec.rb
```diff
@@ -332,7 +332,7 @@
 
     it 'raises "No verification key available" error' do
       token = JWT.encode({}, 'foo')
-      expect { JWT.decode(token, nil, true) }.to raise_error(JWT::DecodeError, 'No verification key available')
```
anakinj, line 335: Did the behaviour change in the test of why was the empty block added? We should be careful when changing old behaviour.

spec/jwt_spec.rb
```diff
@@ -353,6 +353,68 @@
   end
 
   context 'Verify' do
+    context 'find_key' do
```
anakinj, line 356: Would it make sense to have a simple test case also where the keys are given via the second argument to the `JWT.decode` method?

For readability it could make sense be more explicit in what keys are used in the tests. Not maybe relying on the `iss` headers?

spec/jwk/ec_spec.rb
ritikesh: Hey, I did update my branch with the latest commits on master and this test still kept failing. It fails when I run tests on my local machine on the master branch as well. I'm using ruby-2.4.7 and ruby-3.0.1 for running the tests and rubocop.

spec/jwt_spec.rb
ritikesh: > Would it make sense to have a simple test case also where the keys are given via the second argument to the JWT.decode method?

This is already covered, so I didn't want to add redundant tests and use this only for the block versions.

spec/jwt_spec.rb
ritikesh: Nice catch, have made some changes (moved none and no key verification from `Signature#verify` to `Decode#verify_signature`) to ensure there are no breaking behavioural changes.

spec/jwt_spec.rb
ritikesh: > For readability it could make sense be more explicit in what keys are used in the tests. Not maybe relying on the iss headers?

Sorry, I do not understand this... The key is from the original `data` hash. The use-case I'm trying to add tests for is - each ISS can have it's own secret/key. So because I can't pass the secret upfront, because I don't know the issuer, I rely on the `find_key` block to get the issuer from the payload and use the issuer to fetch the secret from my secrets hash. I have added a direct test case as well which does something similar but without the whole issuer thing. The idea remains the same, payload contains information for me, which helps me retrieve the secret/key to verify the JWT. So I can use `find_key` for such use-cases. Let me know if this makes sense.

spec/jwk/ec_spec.rb
anakinj: You're right. I think i fixed it somewhere but apparently not. You'll just have to ignore the failing test for now, sorry for the inconvenience

ritikesh: Alright, reverted that spec change! 👍  Anything else pending for this to get merged? Would you like me to add a changelog & minor version(?) update as well?

lib/jwt/decode.rb
```diff
@@ -33,18 +34,34 @@ def decode_segments
...
 
     def verify_signature
+      return unless @key || @verify
+
+      return if none_algorithm?
+
+      raise JWT::DecodeError, 'No verification key available' unless @key
+
+      Array(@key).each do |key|
+        # return nil to avoid rubocop raising Lint/NonLocalExitFromIterator
```
anakinj, line 44: Can this RuboCop issue be avoided by using something like:
```
return if Array(@key).any?{ |key| verify_signature_for?(key) }
```

spec/jwt_spec.rb
anakinj: Sorry for being unclear, I'll try to rephrase with an example:)

What i tired to say is that the tests are very specific to maybe your use-case, but the functionality is very generic so it can be a little confusing to understand what the tests are really testing.

I think just a set of simple tests testing the array handling would be enough:
``` 
context 'when key given as an array with multiple possible keys' do
      let(:payload) { { 'data' => 'data'} }
      let(:token)   { JWT.encode(payload, secret, 'HS256') }
      let(:secret)  { 'hmac_secret' }
  
      it 'should be able to verify signature when block returns multiple keys' do
        decoded_token = JWT.decode(token, nil, true, { algorithm: 'HS256' }) { |_headers, _payload| ['not_the_secret', secret] }
        expect(decoded_token.first).to eq(payload)
      end

      it 'should be able to verify signature when multiple keys given as a parameter' do
        decoded_token = JWT.decode(token, ['not_the_secret', secret] , true, { algorithm: 'HS256' })
        expect(decoded_token.first).to eq(payload)
      end

      it 'should fail if only invalid keys are given' do
        expect do
          JWT.decode(token, ['not_the_secret', 'not_the_secret_2'] , true, { algorithm: 'HS256' })
        end.to raise_error(JWT::VerificationError, 'Signature verification failed')
      end
    end
```

spec/jwk/ec_spec.rb
anakinj: The changelog and versions are usually updated when a new release is cut. Also the fix for the test is now in #427. Did not dig that deep into this but it seems to be a macOS issue

lib/jwt/decode.rb
ritikesh: great suggestion! done.

spec/jwt_spec.rb
ritikesh: Got it, have add your suggestions! :-)

spec/jwk/ec_spec.rb
ritikesh: Oh okay.

anakinj:
@excpt You have anything to add? I think this is a pretty neat addition. Would be ready to merge.

ritikesh:
Hi @anakinj / @excpt - following up on this. Is there anything pending for this to be merged?

anakinj (approved):

anakinj:
From my part it looks good. Was thinking a second opinion on the feature would be valuable.

ritikesh:
Hi @anakinj / @excpt - circling back on this.

ritikesh:
Hi @excpt - have you had a chance to look at this? Would appreciate your thoughts on how to proceed.

anakinj:
Im thinking we could merge this. Could you still be so kind and rebase to the latest master where the `ruby-head` build should have been fixed.

ritikesh:
Hey, I've rebased my branch to the current master. Thanks.

> Im thinking we could merge this. Could you still be so kind and rebase to the latest master where the `ruby-head` build should have been fixed.

PR #426, opened 2021-06-17
Add documentation for find_key

ritikesh:
NA

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,6 +460,28 @@ rescue JWT::InvalidSubError
 end
 ```
 
+### Finding a Key
+
+To dynamically find the key for verifying the JWT signature, pass a block to the decode block. The block receives headers and the original payload as parameters. It should return with the key to verify the signature that was used to sign the JWT.
+
+```ruby
+issuers = %w[My_Awesome_Company1 My_Awesome_Company2]
+iss_payload = { data: 'data', iss: issuers.first }
+
+secrets = { issuers.first => hmac_secret, issuers.last => 'hmac_secret2' }
+
+token = JWT.encode iss_payload, hmac_secret, 'HS256'
+
+begin
+  # Add iss to the validation to check if the token has been manipulated
+  decoded_token = JWT.decode(token, nil, true, { iss: issuers, verify_iss: true, algorithm: 'HS256' }) do |_headers, payload|
+    secrets[payload['iss']]
+  end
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -460,6 +460,28 @@ rescue JWT::InvalidSubError
...
+iss = 'My_Awesome_Company'
+iss_payload = { data: 'data', iss: iss }
+
+secrets = { iss => hmac_secret }
+
+token = JWT.encode iss_payload, hmac_secret, 'HS256'
+
+begin
+  # Add iss to the validation to check if the token has been manipulated
+  decoded_token = JWT.decode(token, nil, true, { iss: iss, verify_iss: true, algorithm: 'HS256' }) do |_headers, payload|
```
anakinj, line 477: Just a random idea: Would i make sense to have 2 different issuers to pick between in the example.

README.md
```diff
@@ -460,6 +460,28 @@ rescue JWT::InvalidSubError
 end
 ```
 
+### Finding a Key
+
+To dynamically find the secret that was used for signing the JWT, pass a block to the decode block. The block receives headers and the original payload as parameters. It should return with the secret to verify the signature that was used to sign the JWT.
```
anakinj, line 465: To also cover the asymmetric algorithms maybe the wording could be `find the key for verifying the JWT signature`.

The key is not always the secret and not always used for signing tokens.

anakinj (changes requested):
Really good addition to visualize how a block can be used to resolve keys.

A test to verify that the example works in the future would be valuable, some are located in [readme_examples_spec.rb](https://github.com/jwt/ruby-jwt/blob/master/spec/integration/readme_examples_spec.rb)

ritikesh:
> Really good addition to visualize how a block can be used to resolve keys.
> 
> A test to verify that the example works in the future would be valuable, some are located in [readme_examples_spec.rb](https://github.com/jwt/ruby-jwt/blob/master/spec/integration/readme_examples_spec.rb)

Thanks, I've accommodated the suggestions and added the test case. 👍

anakinj (approved):

ritikesh:
Fixed rubocop in https://github.com/jwt/ruby-jwt/commit/8c506e3575e207dd19da25d238dd86481b99d3a3.

anakinj:
Big thanks for the contribution.

PR #431, opened 2021-07-02
verify_claims test shouldnt be within the verify_sub test

andyjdavis:
It appears that the verify_claims tests were within the tests for verify_sub. All I have done is move an 'end' from line to line 247 to line 223 then fixed the indenting.

The tests pass both before and after (and the same number of tests are run).


anakinj:
Thanks for the fix.

PR #437, opened 2021-08-03
Allow regular expressions and procs to verify issuer

rewritten:
The use case is when one has a generic public key resolver (using a block) and there is a need of additional restrictions in a specific context. Of course the rules here are completely arbitrary.

```ruby
# This endpoint will accept JWTs from all registered issuers.
class ApplicationController
  def check_jwt
    token = request.authorization&.delete_prefix('Bearer ')
    JWT.decode(token, nil, true, verify_iss: true, iss: allowed_issuers, algorithm: 'HS256') do |_, claims|
      PublicKeyRegistry.fetch(claims['iss'])
    end
  end

  def allowed_issuers = //
end

# This endpoint will only accept calls from a subset of the registered issuers.
class ChildController < ApplicationController
  # only issuers namespaced on one service
  def allowed_issuers = ->(iss) { iss.start_with('my-internal-service') }
end

# This endpoint will only accept calls from a subset of the registered issuers.
class NewController < ApplicationController
  # only issuers with the current year
  def allowed_issuers = ->(iss) { iss[/\d{4}].to_i == Date.today.year }
end
```


lib/jwt/verify.rb
```diff
@@ -53,9 +53,14 @@ def verify_iss
 
       iss = @payload['iss']
 
-      return if Array(options_iss).map(&:to_s).include?(iss.to_s)
+      options_iss = Array(options_iss).map { |item| item.is_a?(Symbol) ? item.to_s : item }
 
-      raise(JWT::InvalidIssuerError, "Invalid issuer. Expected #{options_iss}, received #{iss || '<none>'}")
+      case iss
+      when *options_iss
```
anakinj, line 59: This is a pretty neat way to handle all the different possible values. Thanks for teaching me :)

README.md
```diff
@@ -370,6 +370,33 @@ rescue JWT::InvalidIssuerError
...
+```ruby
+JWT.decode token, hmac_secret, true,
+           iss: ->(issuer) { issuer.start_with?('My Awesome Company Inc') },
+           verify_iss: true,
+           algorithm: 'HS256'
+```
+
+```ruby
+JWT.decode token, hmac_secret, true,
+           iss: method(:valid_issuer?),
```
anakinj, line 391: does the method definition need to be before calling JWT.decode?

lib/jwt/verify.rb
```diff
@@ -53,9 +53,14 @@ def verify_iss
...
+        return
+      else
+        raise(JWT::InvalidIssuerError, "Invalid issuer. Expected #{options_iss}, received #{iss || '<none>'}")
```
anakinj, line 62: How will this error message look like when there are lambdas and methods in the mix?

lib/jwt/verify.rb
anakinj: This comment can be ignored, feel that I might be overthinking this. Just need to write the behaviour down to be clear:)

Something like this will not work, because we stop calling the lambdas after the first truthy value.

The idea that we require the issuer but some are not allowed.
```
JWT.decode token, hmac_secret, true,
           iss: [->(issuer) { !issuer.blank? },
                 ->(issuer) { issuer != 'The issuer we do not accept' }],
           verify_iss: true,
           algorithm: 'HS256'
```

anakinj: Also supporting EOL rubies is biting this wonderful approach of using a `case`. It does not seem to work with methods

anakinj: Was taking to a few colleagues about this neat way of using a use-case and they pointed out that this works because the `#call` method is aliased to `#===` for procs. So maybe it would be more readable and less magical if the check would just be:

```
 return if options_iss.any? { |option| options === iss }
```

anakinj: What goes for the EOL rubies and the failing test related to passing a method, we could just not run the test with the older rubies until we officially drop the support for the older rubies.

rewritten: > Something like this will not work, because we stop calling the lambdas after the first truthy value.

My understanding is that the list is an "allowlist" (it's the case for strings) so there is no expectation to check further after one positive match.

> the `#call` method is aliased to `#===` for procs

Rubocop complains about any explicit use of `===`. Indeed this PR relies on the behavior of `===` for procs and regular expressions, in addition to the fact that `===` is the same as `==` for string instances. But I tend to agree with BBatsov on this, `===` is an internal tool to implement case matching and pattern matching, which are the user-facing language features to be used. It's like having to use `h = my_object.deconstruct_keys(:a, :b); a, b = h.values_at(:a, :b)` instead of `h => {a: , b:}`.

README.md
rewritten: Method definition order does not matter, they are all available when an instance is created.

lib/jwt/verify.rb
rewritten: It will be something like `"Invalid issuer. Expected #Proc<...>, received foo_bar"`. Not the nicest case but there is no nice way to print out a proc. I also did not want to make extra cases.

If one wants a nicer output, they can define `to_s` on their proc:

```rb
matcher = ->(issuer) { issuer.start_with?('foo') && issuer != 'foo_bar' }
def matcher.to_s
  'A nicer description'
end
```

lib/jwt/verify.rb
rewritten: > What goes for the EOL rubies and the failing test related to passing a method

I will try to make them pass.

README.md
```diff
@@ -370,6 +370,33 @@ rescue JWT::InvalidIssuerError
...
+           verify_iss: true,
+           algorithm: 'HS256'
+
+def valid_issuer?(issuer)
```
rewritten, line 395:
```suggestion
# somewhere in the same class:
def valid_issuer?(issuer)
```

spec/jwt/verify_spec.rb
```diff
@@ -145,6 +145,65 @@
...
+      it 'must raise JWT::InvalidIssuerError when the payload does not include an issuer' do
+        expect do
+          described_class.verify_iss(base_payload, options.merge(iss: method(:issuer_start_with_first?)))
+        end.to raise_error(JWT::InvalidIssuerError, /received <none>/)
+      end
+
+      it 'must allow a method that returns true to pass' do
+        described_class.verify_iss(payload, options.merge(iss: method(:issuer_start_with_ruby?)))
+      end
+    end
```
rewritten, line 206:
```suggestion
    end if Method.instance_methods.include?(:===)
```

lib/jwt/verify.rb
rewritten: > I will try to make them pass.

No way - just skipped them

anakinj: You make a fair point, let's have it like this. I still think calling the explicit `#===` would be less code and a little more readable.

README.md
anakinj: If this would have been a script or a snipped pasted into a console i think the order would matter.

anakinj:
@rewritten Are you planning on continuing on the feature. Just trying to ignore the tests on the old rubies would be enough to get the workflow happy. 

Also thinking that after the next release we will drop the official support fo EOLed Rubies and this PR can be merged without changes.

rewritten:
@anakinj owww I completely missed your message. I will keep this PR updated to the latest master and make it work on non-EOLed rubies (and therefore I don't have to dance around them)

anakinj:
@rewritten EOL ruby support is now dropped from the master branch. Think a rebase will make everything greener.

rewritten:
I'll rebase this week

rewritten:
@anakinj  ready for CI

rewritten:
Should be fixed now - I have run rubocop locally and it does not complain anymore

anakinj (approved):

anakinj:
Thanks for your contribution @rewritten. Highly appreciated.

PR #439, opened 2021-08-19
Raise a JWT::DecodeError when token is not a String

kalilz4485:
Hello,

Currently when doing

```ruby
JWT.decode(nil, nil)
```
You get a `JWT::DecodeError (Nil JSON web token)`

or 

```ruby
JWT.decode('invalid', nil)
```
You get a `JWT::DecodeError (Not enough or too many segments)`

But we don't check for anything else than nil, everything else will supposedly fail at the `.split` in the `initialize`
e.g.
```ruby
JWT.decode(10, nil)
```
will give `NoMethodError (undefined method 'split' for 3:Integer)`

The only question is should this be the gem's responsibility to check that ? And if yes should we do the same with the secret (gives a `TypeError` which is slightly better)

PR is as close as possible from previous code to return a `JWT::DecodeError`

### DIFF
--- a/lib/jwt/decode.rb
+++ b/lib/jwt/decode.rb
@@ -9,7 +9,7 @@ module JWT
   # Decoding logic for JWT
   class Decode
     def initialize(jwt, key, verify, options, &keyfinder)
-      raise(JWT::DecodeError, 'Nil JSON web token') unless jwt
+      raise(JWT::DecodeError, "#{jwt.class} JSON web token") unless jwt.class == String
       @jwt = jwt
       @key = key
       @options = options
--- a/spec/jwt_spec.rb
+++ b/spec/jwt_spec.rb
@@ -438,6 +438,13 @@ RSpec.describe JWT do
     end
   end
 
+  context 'when token is not a String' do
+    it 'raises JWT::DecodeError' do
+      expect { JWT.decode(nil, nil, true) }.to raise_error(JWT::DecodeError, 'NilClass JSON web token')
+      expect { JWT.decode(1, nil, true) }.to raise_error(JWT::DecodeError, 'Integer JSON web token')
+    end
+  end
+
   context 'a token with no segments' do
     it 'raises JWT::DecodeError' do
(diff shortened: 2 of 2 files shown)


lib/jwt/decode.rb
```diff
@@ -9,7 +9,7 @@ module JWT
   # Decoding logic for JWT
   class Decode
     def initialize(jwt, key, verify, options, &keyfinder)
-      raise(JWT::DecodeError, 'Nil JSON web token') unless jwt
+      raise(JWT::DecodeError, "#{jwt.class} JSON web token") unless jwt.class == String
```
anakinj, line 12: Would suggest using the `Object#is_a?`to check the type of the given parameter. The method also takes inheritance into consideration.

Also think it would solve the sourceleve-bot whining.

anakinj (changes requested):

lib/jwt/decode.rb
anakinj: Im wondering if this could be just a `raise TypeError, 'JSON web token is expected to be a String #{jwt.class} given' unless jwt.is_a?(String)`


Maybe to not brake backwards compatibility:
`raise JWT::DecodeError, 'JSON web token is expected to be a String #{jwt.class} given' unless jwt.is_a?(String)`

spec/jwt_spec.rb
```diff
@@ -438,6 +438,13 @@
     end
   end
 
+  context 'when token is not a String' do
+    it 'raises JWT::DecodeError' do
+      expect { JWT.decode(nil, nil, true) }.to raise_error(JWT::DecodeError, 'NilClass JSON web token')
+      expect { JWT.decode(1, nil, true) }.to raise_error(JWT::DecodeError, 'Integer JSON web token')
```
anakinj, line 444: Integer is still Fixnum in some older rubies that the gem supports (See the failing tests)

a0960909060:
從我的iPhone 傳送

anakinj:
I'm going to close this one as stale. Please reopen or create a new one if work is still going to happen related to this.

PR #442, opened 2021-08-21
Remove context specific code from security utils

anakinj:
The `JWT::SecurityUtils` module has over the years been used as a placeholder for context specific methods.

This PR moves the "utility" methods into the context they actually are used from.

Also updated the SecureUtils.secure_compare to take advantage of the latest optimisations available.


anakinj:
Im going to close this. Will make a new attempt from the current head.

PR #341, opened 2019-11-26
Add more docs about JWK support

take:
docs for the following code

- https://github.com/jwt/ruby-jwt/commit/922dac620bf1e707f36ee35eef1e5881d1e32cc5#diff-35b558119ace51c78f5563ad9f4d2ec3R8
- https://github.com/jwt/ruby-jwt/commit/922dac620bf1e707f36ee35eef1e5881d1e32cc5#diff-35b558119ace51c78f5563ad9f4d2ec3R49

PR #443, opened 2021-09-08
fix document about passing JWKs as a simple Hash

takayamaki:
Hi.

I found a gap between documentation and implementation.

document: https://github.com/jwt/ruby-jwt/pull/341
implementation: https://github.com/jwt/ruby-jwt/pull/348

### DIFF
--- a/README.md
+++ b/README.md
@@ -518,7 +518,7 @@ end
 or by passing JWK as a simple Hash
 
 ```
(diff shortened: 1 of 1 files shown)


anakinj:
Thanks for noticing and fixing.

anakinj (approved):

PR #445, opened 2021-09-21
Pass kid param through JWT::JWK.create_from

shaun-guth-allscripts:
The underlying keypair classes already support an optional second parameter for the key id; however, the JWT::JWK interface didn't provide any mechanism for setting them via the constructor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -495,7 +495,7 @@ JWT.decode token, hmac_secret, true, { required_claims: ['exp'], algorithm: 'HS2
 JWK is a JSON structure representing a cryptographic key. Currently only supports RSA public keys.
 
 ```ruby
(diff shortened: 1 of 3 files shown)


anakinj:
Looks great. Could you be so kind and fix the RuboCop issue mentioned here: https://github.com/jwt/ruby-jwt/pull/445/checks?check_run_id=3740502864

shaun-guth-allscripts:
> Looks great. Could you be so kind and fix the RuboCop issue mentioned here: https://github.com/jwt/ruby-jwt/pull/445/checks?check_run_id=3740502864

This should be fixed now - thanks for looking into it!

anakinj (approved):

PR #446, opened 2021-09-27
feat(EdDSA): Accept EdDSA as algorithm header

Pierre-Michard:
Some implementations of EdDSA algorithm use 'EdDSA' as header instead of ED25519. This MR aims to make this library compatible with such clients.


anakinj:
Looks great!  Think this addresses a part of #334 also

anakinj (approved):

PR #454, opened 2021-12-13
Use Ruby built-in url-safe base64 methods

bdewater:
No duplicated code and some small [performance improvements from upstream](https://github.com/ruby/ruby/commits/master/lib/base64.rb)


lib/jwt/decode.rb
```diff
@@ -128,8 +128,8 @@ def signing_input
     end
 
     def parse_and_decode(segment)
-      JWT::JSON.parse(JWT::Base64.url_decode(segment))
-    rescue ::JSON::ParserError
+      JWT::JSON.parse(Base64.urlsafe_decode64(segment))
+    rescue ::JSON::ParserError, ArgumentError
```
anakinj, line 132: What was the case where the ArgumentError exception was raised?

anakinj (commented):
Im liking it...

excpt (approved):

lib/jwt/decode.rb
anakinj: Think i've figured it out. It was when the segment is not valid base64

PR #464, opened 2022-01-04
Bring back Ruby 2.5 support and CodeClimate coverage reports

anakinj:
Think we should take one step back in dropping older Ruby support, keeping the EOLed 2.5 support until it adds some problems for us. The gem works just fine with 2.5, no new fancy feature taken into use.

Also this PR brings back CodeClimate coverage reports.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,27 @@
+---
+name: coverage
+on:
+  push:
+    branches:
+      - "master"
+jobs:
+ coverage:
+    name: coverage
+    runs-on: ubuntu-20.04
+    env:
+      BUNDLE_GEMFILE: 'gemfiles/rbnacl.gemfile'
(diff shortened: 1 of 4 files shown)


anakinj:
Also we should decide what to do with the `CC_TEST_REPORTER_ID`. Continue to keep it public, with the risk of anyone could push invalid coverage reports to the CodeClimate ruby-jwt project.

If we want to make it private we need to rotate it and add it to the project secrets and coverage runs will only be executed when PRs are merged to master.

The value in this PR was fetched from some old Travis CI log from an old coverage run.

anakinj:
Im going to merge this to see if we can get the coverage report moving. If we decide to still drop 2.5 we can do that later.

anakinj:
Apparently I had access to the CodeClimate repository. I've now rotated the CodeClimate TEST REPORTER ID. 

So for now the coverage reports are only for master build. Could develop it a litte further to have some feedback on the PRs if the coverage goes down.

PR #476, opened 2022-04-25
Fix RuboCop TODOs

typhoon2099:
Mostly autofixes in here, plus the addition of a super to `JWT::JWK::KeyBase.inherited`.

### DIFF
--- a/lib/jwt/jwk/key_base.rb
+++ b/lib/jwt/jwk/key_base.rb
@@ -11,6 +11,7 @@ module JWT
       end
 
       def self.inherited(klass)
+        super
         ::JWT::JWK.classes << klass
       end
     end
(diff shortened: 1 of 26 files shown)


lib/jwt/jwk/key_base.rb
```diff
@@ -11,6 +11,7 @@ def initialize(keypair, kid = nil)
       end
 
       def self.inherited(klass)
+        super
```
anakinj, line 14: Dunno what the guidelines say, but feels like super should be the last thing happening in this method.

anakinj:
I really like this change. I have a tiny-little-supersmall concern about the addition of `# frozen_string_literal: true`. Did not check that deeply into what kind of strings were used in the files but it could potentially break someones implementation if they for some strange reason were mutating the strings coming out of this gem.

On the other hand the next version will also drop support for older rubies so maybe now is a good time to add that to everywhere.

anakinj (approved):

PR #479, opened 2022-05-07
Add ES256K support

jimryan:
This adds support for the ES256K algorithm (secp256k1 curve). Not sure if this is something that you'd like to support, but I needed it, so figured I'd open a PR. It's useful for blockchain applications.

I'm far from a cryptography expert, so please do carefully review this. It seems to work for me, though.


anakinj:
Thanks. Great addition.


Rubocop is apparently complaining about long file names. I think it would be time to refactor the `lib/jwt/jwk/ec.rb` a little. But that could be addressed in another PR. For now i guess bumping the max values in `.rubocop.yml` a little would be enough.

lib/jwt/algos/ecdsa.rb
```diff
@@ -45,7 +46,7 @@ def curve_by_name(name)
 
         {
           algorithm: algorithm,
-          digest: algorithm.sub('ES', 'sha')
+          digest: algorithm.sub('ES', 'sha').sub('K', '')
```
anakinj, line 49: Would love to see the algorithm being explicit in the `NAMED_CURVES` constant instead of trying to preserve the numbers in the name. But that could be addressed in another PR.

jimryan: Totally agree. That stood out to me as well. I'd be happy to include that change in this PR as I do think the "K" substitution pushes the current approach too far.

jimryan:
Thanks for the super quick review @anakinj! I'll bump that rubocop value.

lib/jwt/algos/ecdsa.rb
anakinj: That sounds super, just go for it:)

jimryan:
@anakinj All set! Thanks for the feedback. I really wanted to refactor how curve aliases are defined (i.e. secp256r1), but decided that'd be best handled in a separate PR.

anakinj:
Until there are more aliases I would not complicate the definitions. But when you say it I got curious, what kind of approach where you thinking on, my opinions are not that strong:).

Really like this. Thanks for your effort.

anakinj (approved):

PR #480, opened 2022-05-29
Fixed a few style issues

anakinj:
This PR:

- Remove the need for the `.rubocop_todo.yml`
- Configures SourceLevel to use a more modern RuboCop version. The latest version they support seems to be 1.9.1 so we'll go with that.
- Start running RuboCop on Ruby 3.0 for the CI (3.1 is echoing a warning)
- Removes a few disabled cops, that were trivial to fix and some were not doing any effect at all.
- Got rid of the dependency to https://github.com/excpt/linters (reek config moved to the repo instead)


excpt (approved):
💯 

Thank you!

Issue #474, opened 2022-03-11
Use RFC 7638 JWK Thumbprints as kid


bellebaum:
The current method for calculating JWK `kid`s is something like

```ruby
      def generate_kid(ec_keypair)
        _crv, x_octets, y_octets = keypair_components(ec_keypair)
        sequence = OpenSSL::ASN1::Sequence([OpenSSL::ASN1::Integer.new(OpenSSL::BN.new(x_octets, BINARY)),
                                            OpenSSL::ASN1::Integer.new(OpenSSL::BN.new(y_octets, BINARY))])
        OpenSSL::Digest::SHA256.hexdigest(sequence.to_der)
      end
```

for EC and something similar for RSA. I am unsure if there is some standard that this behavior is based on.

[RFC 7638](https://datatracker.ietf.org/doc/html/rfc7638) describes a canonical way to derive a Thumbprint for a JWK,
the `base64url` encoding of which is advertised as a "good" `kid` is the following:

> 1.  Construct a JSON object [[RFC7159](https://datatracker.ietf.org/doc/html/rfc7159)] containing only the required
>    members of a JWK representing the key and with no whitespace or
>    line breaks before or after any syntactic elements and with the
>   required members ordered lexicographically by the Unicode
>       [[UNICODE](https://datatracker.ietf.org/doc/html/rfc7638#ref-UNICODE)] code points of the member names.  (This JSON object is
>       itself a legal JWK representation of the key.)
>
>   2.  Hash the octets of the UTF-8 representation of this JSON object
>       with a cryptographic hash function H.  For example, SHA-256 [[SHS](https://datatracker.ietf.org/doc/html/rfc7638#ref-SHS)]
>       might be used as H.  See [Section 3.4](https://datatracker.ietf.org/doc/html/rfc7638#section-3.4) for a discussion on the
>       choice of hash function.

In some (admittedly rare) cases algorithms seem to rely on this behavior, which is why many popular JWT libraries seem to implement it (including `json-jwt`).
Should this be implemented here as well?

bellebaum:
I should mention that the above current method should result in the same kid for keys with the same coordinates in different curves. I am unsure whether or not that can be a security risk.

anakinj:
You're right. The kid calculation is pretty random at the moment. Problem with it is that it's "out there" already, so changing it will be somewhat challenging without breaking things.

Think we could at least support calculating the thumbprints for the jwks as a first step.

[...]

PR #481, opened 2022-05-29
Support JWK thumbprints as key ids

anakinj:
This addresses #474.

Allow the default and custom kid generation to be changed to a more "standard" way of generating JWK kids.

Also introduces a configuration concept via`::JWK.configuration` where the default values has been moved.

Would very much like to deprecate the current way of kid generation, but we would need to notify about the deprecation in some way before that.


excpt (changes requested):
Please update the version.rb to 3.0.0.dev

It's an API breaking change. 

Great work! 👍

anakinj:
The API should not have changed in this case, the intention was that at least. What API do you have on your mind?

anakinj:
@excpt got very curious on what im missing:)

excpt:
I see that I have chosen my word poorly. 😖 

Clarification:

You want to depricate the current used way of kid generation.

This will break existing implementations while continuing using the 2.x versions of this gem.

I recommend bumping the version to 3.0 in that case. So this is will be clear for all the direct and mostly indirect users of this gem.

This allows us to continue the development and it's still possible to continue supporting the 2.x versions.

anakinj:
Thanks for the clarification 👍 . I will in my turn clarify my thoughts:

- The changes are not changing the current behaviour at all so it's possible to ship in a 2.x version with optional support for this.
- If someone would like to use the thumbprint as the kid the gem can be configured with `JWT.configuration.jwk.kid_generator_type = :rfc7638_thumbprint`. This will change the behaviour of the kid generation to generate the new kind of kids. The default value for this will be `JWT.configuration.jwk.kid_generator_type = :custom`
- In a future 2.x version we could add a deprecation warning on load that the defaults will change and explain what it means.

anakinj:
@excpt ok to merge? No changes to the current behaviour and deprecation will happen at a later stage with some kind of warnings to the user.

excpt (approved):
Thank you! 👍

Issue #483, opened 2022-06-07
Regression: 2.4.0 results in an ArgumentError with an invalid Base64 string (expecting JWT::DecodeError)


freakyfelt:
Calling JWT.decode(token) used to result in a JWT::DecodeError, however this is now raising an `ArgumentError` that must be caught separately

```
➜  ~ gem install jwt '= 2.4.0'
Successfully installed jwt-2.4.0
Parsing documentation for jwt-2.4.0
Installing ri documentation for jwt-2.4.0
Done installing documentation for jwt after 0 seconds
➜  ~ irb
3.1.1 :003 > JWT.decode('hello.there.world')
3.1.1 :001 > require 'jwt'
3.1.1 :002 > JWT.decode('hello')
~/.rvm/gems/ruby-3.1.1/gems/jwt-2.4.0/lib/jwt/decode.rb:108:in `validate_segment_count!': Not enough or too many segments (JWT::DecodeError)
	from ~/.rvm/gems/ruby-3.1.1/gems/jwt-2.4.0/lib/jwt/decode.rb:25:in `decode_segments'
	from ~/.rvm/gems/ruby-3.1.1/gems/jwt-2.4.0/lib/jwt.rb:28:in `decode'
	from (irb):2:in `<main>'
	from ~/.rvm/rubies/ruby-3.1.1/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
	from ~/.rvm/rubies/ruby-3.1.1/bin/irb:25:in `load'
	from ~/.rvm/rubies/ruby-3.1.1/bin/irb:25:in `<main>'
3.1.1 :003 > JWT.decode('hello.there.world')
<internal:pack>:309:in `unpack1': invalid base64 (ArgumentError)
	from ~/.rvm/rubies/ruby-3.1.1/lib/ruby/3.1.0/base64.rb:74:in `strict_decode64'
	from ~/.rvm/rubies/ruby-3.1.1/lib/ruby/3.1.0/base64.rb:108:in `urlsafe_decode64'
	from ~/.rvm/gems/ruby-3.1.1/gems/jwt-2.4.0/lib/jwt/decode.rb:120:in `decode_crypto'
	from ~/.rvm/gems/ruby-3.1.1/gems/jwt-2.4.0/lib/jwt/decode.rb:27:in `decode_segments'
[...]

PR #484, opened 2022-06-07
fix(JWT::decode): raise JWT::DecodeError on invalid signature

freakyfelt:
Resolves https://github.com/jwt/ruby-jwt/issues/483


anakinj (approved):

anakinj:
Thanks for the quick fix.

Issue #487, opened 2022-06-07
JWT.decode broken in v2.4: `unpack1': invalid base64 (ArgumentError)


collimarco:
The update from v2.3 to v2.4 of this gem breaks our gem.

This is the error:

```
<internal:pack>:281:in `unpack1': invalid base64 (ArgumentError)
	from /Users/collimarco/.rbenv/versions/3.0.4/lib/ruby/3.0.0/base64.rb:74:in `strict_decode64'
	from /Users/collimarco/.rbenv/versions/3.0.4/lib/ruby/3.0.0/base64.rb:106:in `urlsafe_decode64'
	from /Users/collimarco/.rbenv/versions/3.0.4/lib/ruby/gems/3.0.0/gems/jwt-2.4.0/lib/jwt/decode.rb:120:in `decode_crypto'
	from /Users/collimarco/.rbenv/versions/3.0.4/lib/ruby/gems/3.0.0/gems/jwt-2.4.0/lib/jwt/decode.rb:27:in `decode_segments'
	from /Users/collimarco/.rbenv/versions/3.0.4/lib/ruby/gems/3.0.0/gems/jwt-2.4.0/lib/jwt.rb:28:in `decode'
	from /Users/collimarco/Sites/cuber-gem/lib/cuber/cuberfile_validator.rb:132:in `validate_key'
	from /Users/collimarco/Sites/cuber-gem/lib/cuber/cuberfile_validator.rb:26:in `validate'
	from /Users/collimarco/Sites/cuber-gem/lib/cuber/cli.rb:43:in `validate_cuberfile'
	from /Users/collimarco/Sites/cuber-gem/lib/cuber/cli.rb:20:in `initialize'
	from /Users/collimarco/Sites/cuber-gem/bin/cuber:3:in `new'
	from /Users/collimarco/Sites/cuber-gem/bin/cuber:3:in `<top (required)>'
	from /Users/collimarco/.rbenv/versions/3.0.4/lib/ruby/gems/3.0.0/bin/cuber:25:in `load'
	from /Users/collimarco/.rbenv/versions/3.0.4/lib/ruby/gems/3.0.0/bin/cuber:25:in `<top (required)>'
	from /Users/collimarco/.rbenv/versions/3.0.4/lib/ruby/3.0.0/bundler/cli/exec.rb:58:in `load'
[...]

PR #489, opened 2022-06-08
Test for decode behaviour with non-alphabetic chars in base64

anakinj:
Added a test to document the behaviour of the decode method when tokens including some non-alphabetic chars are given.


anakinj:
Im going to merge this. Think this is the behaviour that is causing #487

PR #491, opened 2022-06-21
Rescue RbNaCl exception for EdDSA wrong key

### LINKED ISSUE #490: Rotating keys with ED25519
I'm trying to implement rotating keys, I'm not sure if I'm doing it right.

```ruby
payload = JWT.decode(token, nil, true, { algorithm: 'ED25519' }) do |_headers, payload|
  [old_public_key, public_key]
end
=> raise 'signature was forged/corrupt (RbNaCl::BadSignatureError)'
```

`decode` is supposed to test all keys with `return if Array(@key).any? { |key| verify_signature_for?(key) }` but `verify_signature_for?` will raise an error if the key is invalid instead of returning `false`.

Is it the expected behavior? If not I can submit a PR. If it is the intended behavior how should I implement rotating keys?

Thank you!

n-studio:
Fixes #490 

Instead of raising an error when a public key is invalid, we should return `false` so the key finder can test the next key.

### DIFF
--- a/spec/jwt_spec.rb
+++ b/spec/jwt_spec.rb
@@ -200,7 +200,7 @@ RSpec.describe JWT do
           data[alg] = JWT.encode payload, data["#{alg}_private"], alg
         end
 
-        let(:wrong_key) { OpenSSL::PKey.read(File.read(File.join(CERT_PATH, 'ec256-wrong-public.pem'))) }
+        let(:wrong_key) { RbNaCl::Signatures::Ed25519::SigningKey.generate.verify_key }
 
         it 'should generate a valid token' do
           jwt_payload, header = JWT.decode data[alg], data["#{alg}_public"], true, algorithm: alg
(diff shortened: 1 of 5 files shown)


anakinj (changes requested):

n-studio:
@anakinj Thank you for the review! I updated my code accordingly.

spec/jwt_spec.rb
```diff
@@ -218,7 +218,7 @@
 
         it 'wrong key should raise JWT::DecodeError' do
           expect do
-            JWT.decode data[alg], wrong_key
+            JWT.decode data[alg], wrong_key, true, algorithm: alg
```
anakinj, line 221: Oh, this was a good catch, the test has been wrong all the time, right. Like passing for the wrong reasons.

anakinj (approved):

anakinj:
This is great stuff. Thank you for the fix.

PR #494, opened 2022-06-22
Improved RSA signing and verification

anakinj:
Some improvements for the RSA algorithm:
- One step closer to get rid of the algorithm specific methods from `::JWT:SecurityUtils` (Something that was started on #442)
- Stricter requirements for the RSA key on both signing and verification. 
- Simpler resolving of the digests per algorithm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 **Fixes and enhancements:**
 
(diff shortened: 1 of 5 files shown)


anakinj:
#512 will do some adjustments to how the algorithm classes are structured

Issue #495, opened 2022-06-22
Use OpenSSL 3


guillaumewrobel:
[Heroku-22 stack](https://devcenter.heroku.com/articles/heroku-22-stack#openssl-3) has been released, and it does not support OpenSSL 1.1 but 3.0.

The `ruby-jwt` broke with the following error

> rsa#set_key= is incompatible with OpenSSL 3.0

`ruby-openssl`v3 [has been released](https://github.com/ruby/openssl/releases/tag/v3.0.0) and should be used

anakinj:
Hi, just curious what version of the gem you are using? There was one fix for compatibility with OpenSSL 3.0 released in 2.4.0

Also this error reminds me of something that was fixed in #333

ptiseo-tegria:
Just hit this problem with Ubuntu 22.04.1LTS and jwt 2.4.1

anakinj:
Thanks for the report. 

Some kind of support for OpenSSL 3.0 will be shipped in the next release of this gem, would be great if someone would try the current master version in a real-world example.

ptiseo-tegria:
@anakinj maybe I didn't understand, but my issue is a real-world codebase with the 2.4.1 master release? How else can I help?

anakinj:
Sorry, was a little unclear. Current master being what is on GitHub that will eventually be something greater than 2.4.1, most probably 2.5.0.

The master branch very often contains features and fixes that has not been released yet, as in this case. So for example by looking at the differences from 2.4.1 master you can [compare the v2.4.1 tag with master](https://github.com/jwt/ruby-jwt/compare/v2.4.1...master). 

And by trying it out in a real-world example would mean to use the actual master branch of this repo in your codebase, specifying in your gemfile:

```
  gem 'jwt', github: 'jwt/ruby-jwt'
```

ptiseo-tegria:
Np. Glad to assist. Made the change and got a different error:

```
/usr/share/rvm/gems/ruby-3.1.2/bundler/gems/ruby-jwt-0e7dcfacd822/lib/jwt/jwk/ec.rb:124:in `singleton class': undefined method `openssl_3?' for JWT:Module (NoMethodError)

        if ::JWT.openssl_3?
                ^^^^^^^^^^^
        from /usr/share/rvm/gems/ruby-3.1.2/bundler/gems/ruby-jwt-0e7dcfacd822/lib/jwt/jwk/ec.rb:92:in `<class:EC>'
[...]

PR #496, opened 2022-06-23
Support OpenSSL >= 3.0

anakinj:
OpenSSL 3.0 is taking over and that brings some compatibility challenges to gems like this. Related to #495.

https://github.com/ruby/openssl/blob/master/History.md#compatibility-notes

The change that has the most impact on the ruby-jwt gem is that the key and keypairs that are created using openssl is not mutable anymore. Before the Ruby openssl gem provides some nicer API to generate them based on the raw parameters we need to generate the ASN.1 sequences and have them as the base for creating the key objects.


anakinj:
The behaviour is different when passing an empty secret to the HMAC signing because of https://github.com/openssl/openssl/issues/13089

Just marked the tests as pending for openssl 3

anakinj:
Im going to trust the tests and merge this one. It works at least better than the current version...

PR #502, opened 2022-07-15
Fix NoMethodError when a 2 segment token is missing 'alg' header

cmrd-senya:
This PR fixes an issue that was introduced at https://github.com/jwt/ruby-jwt/pull/425

Currently if one passes a token of 2 segments without `alg` header the `decode` method raises `NoMethorError` because `algorithm` is `nil` but it tries to call `casecmp` on it here:
https://github.com/jwt/ruby-jwt/blob/master/lib/jwt/decode.rb#L116

The expected result would be to raise an exception that is inherited from `JWT::DecodeError` instead of a generic exception like `NoMethodError`.

### DIFF
--- a/lib/jwt/decode.rb
+++ b/lib/jwt/decode.rb
@@ -113,7 +113,7 @@ module JWT
     end
 
     def none_algorithm?
-      algorithm.casecmp('none').zero?
+      algorithm == 'none'
     end
 
     def decode_crypto
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Bring back the old Base64 (RFC2045) deocode mechanisms [#488](https://github.com/jwt/ruby-jwt/pull/488) ([@anakinj](https://github.com/anakinj)).
 - Rescue RbNaCl exception for EdDSA wrong key [#491](https://github.com/jwt/ruby-jwt/pull/491) ([@n-studio](https://github.com/n-studio)).
 - New parameter name for cases when kid is not found using JWK key loader proc [#501](https://github.com/jwt/ruby-jwt/pull/501) ([@anakinj](https://github.com/anakinj)).
+- Fix NoMethodError when a 2 segment token is missing 'alg' header [#502](https://github.com/jwt/ruby-jwt/pull/502) ([@cmrd-senya](https://github.com/cmrd-senya)).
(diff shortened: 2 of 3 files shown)


lib/jwt/decode.rb
```diff
@@ -113,6 +113,8 @@ def segment_length
     end
 
     def none_algorithm?
+      return false unless algorithm.respond_to? :casecmp
```
anakinj, line 116: I wonder if it would just be simpler to check if it's a string or just is truthy.

```
return false unless algorithm
```

```
return false unless algithm.is_a?(String)
```

anakinj (changes requested):

anakinj:
Great catch and nice with additional tests to ensure this behaviour in the future.

Added a comment about the added condition. Would suggest making the check more generic to avoid checking for a certain method.

lib/jwt/decode.rb
anakinj: Maybe even the original approach would be the simplest.

`algorithm == 'none'`

cmrd-senya: Checking for a specific type is not very Ruby-way approach. Because Ruby promotes duck typing, and therefore it shouldn't matter if a class is directly based on a `String` or it implements exactly the same interface independently - the code should work anyways. So checking for the presence of a certain method is quite common in Ruby to ensure the interface of an object is supported, I think it is intentionally preferred over checking the actual class.

Checking `algorithm == 'none'` is fine, the only thing is that it stops supporting case-insensitive matches, but I have quickly checked the RFC and it never explicitly requires to support anything other than `none`.

I have also found that algorithm name is supposed to be case sensitive. 

image

And `none` is listed as an algorithm name:

image

 So I guess supporting something like `None` or `nOnE` could be seen as a spec violation.

https://datatracker.ietf.org/doc/html/rfc7518#section-7.1.1

cmrd-senya:
@anakinj updated

anakinj:
Looks great. Would you be so kind and add a changelog entry to the `CHANGELOG.md` under the fixes?

cmrd-senya:
@anakinj done

anakinj (approved):

anakinj:
Super! Thank you for the effort fixing this!❤️

PR #506, opened 2022-08-24
Release v2.5.0

anakinj:
Here is the preparations for the next release.


anakinj:
@excpt H

anakinj:
@excpt here is a PR with the preparations for v2.5.0

PR #510, opened 2022-09-03
Allow algorithm to be a class

anakinj:
This PR improves the algorithm handling on both encoding and decoding:

- Support for passing custom algorithms to handle the signing and verification of tokens
- When decoding the role of the `alg` header is minimized to only be required and value used to sort the given algorithm array.


anakinj:
I'll cleanup things and come back to this in another PR.

PR #512, opened 2022-10-01
Support algorithm to be given as a class

anakinj:
This PR improves the algorithm handling on both encoding and decoding:

- Support for passing custom algorithms to handle the signing and verification of tokens
- When decoding the alg header is not used as anything else than a hint for the algorithm to try first from the allowed algorithms.


anakinj:
Would you @excpt have time to check this out. Would enable some further improvements to be made to the signing and verification.

lib/jwt/algos/algo_wrapper.rb
```diff
@@ -0,0 +1,30 @@
...
+    class AlgoWrapper
+      attr_reader :alg
+
+      def initialize(alg, cls)
+        @alg = alg
+        @cls = cls
+      end
+
+      def valid_alg?(alg)
+        alg && @alg.casecmp(alg).zero?
```
MatteoPierro, line 14: Probably this can be simplified in this way `@alg.casecmp(alg)&.zero?`

lib/jwt/algos/algo_wrapper.rb
```diff
@@ -0,0 +1,30 @@
...
+      end
+
+      def sign(data:, signing_key:)
+        @cls.sign(@alg, data, signing_key)
```
MatteoPierro, line 18: here the accessor for alg can be used

spec/jwt_spec.rb
```diff
@@ -771,4 +771,75 @@
...
+        def valid_alg?(alg)
+          alg == self.alg
+        end
+      end
+    end
+
+    it 'uses whatever is the implementation' do
+      token = JWT.encode(payload, 'secret', custom_algorithm)
+      expect(token).to eq('eyJhbGciOiJjdXN0b20ifQ.eyJ1c2VyX2lkIjoic29tZUB1c2VyLnRsZCJ9.Y3VzdG9tX3NpZ25hdHVyZQ')
+      expect(JWT.decode(token, 'secret', true, algorithm: custom_algorithm))
```
MatteoPierro, line 810: What is actually expected here? Maybe the `payload`?

spec/jwt_spec.rb
```diff
@@ -771,4 +771,75 @@
...
+        def valid_alg?(alg)
+          alg == self.alg
+        end
+      end
+    end
+
+    it 'uses tries until something matches' do
+      token = JWT.encode(payload, 'secret', custom_algorithm.new('custom_signature'))
+      expect(token).to eq('eyJhbGciOiJjdXN0b20ifQ.eyJ1c2VyX2lkIjoic29tZUB1c2VyLnRsZCJ9.Y3VzdG9tX3NpZ25hdHVyZQ')
+      expect(JWT.decode(token, 'secret', true, algorithms: [custom_algorithm.new('not_this'), custom_algorithm.new('custom_signature')]))
```
MatteoPierro, line 842: same as https://github.com/jwt/ruby-jwt/pull/512/files#r988237828

spec/jwt_spec.rb
```diff
@@ -771,4 +771,75 @@
...
+
+  context 'when algorithm is a custom class and multiple is given' do
+    let(:custom_algorithm) do
+      Class.new do
+        def initialize(signature)
+          @signature = signature
+        end
+
+        def sign(*)
+          'custom_signature'
```
MatteoPierro, line 822: here should be `@signature`, right?

spec/jwt_spec.rb
anakinj: Very good question. A set of new improved tests pushed.

lib/jwt/algos/algo_wrapper.rb
anakinj: Good call. I went even one step further and did a `...zero? == true` to ensure the return value always is a boolean

MatteoPierro (approved):
Looks great now 👍

Issue #518, opened 2022-10-06
JWK Parameters


bellebaum:
Hey there :)

I was trying to use this gem, and the JWK implementation in particular, to handle any kind of cryptographic material for which I would normally go with several files in a format like PEM. This includes public and private key pairs, as well as certificates.
In principle, the JWK specs are built to handle all this kind of stuff, as can be seen e.g. in the [IANA registry](https://www.iana.org/assignments/jose/jose.xhtml#web-key-parameters).
Note in particular the `x5*` claims for certificates.

It would be nice if there were some option to include these parameters in `JWT::JWK` and have it be exported alongside the actual key values. Currently this is not the case, because most values (excluding `kid` and `kty`) are stored internally in an OpenSSL-compatible format (which is a good thing, I think, since cryptographic operations are often a lot more time-critical than serialization).

I do recognize that supporting the full set of semantics connected to these additional parameters may be a massive increase in scope for this project (e.g. any certificates in `x5c` must be signed by the next one in the array (let's not start talking about conformance to all X.509v3 extensions limiting who can sign what) with the first certificate referencing the very key in the JWK, and the `use` parameter should probably also be checked at some point or another).
A much more reasonable approach, if properly documented, would be to just allow people to add custom properties to the JWK. This is in a way similar to how this gem supports adding arbitrary claims to a JWT, but only actively provides guidance and validation for some of them (e.g. `iss`, `aud`, etc.).

The way I would envision this is by using the `options` argument in [KeyBase](https://github.com/jwt/ruby-jwt/blob/9b1a2eac9623a74669e2361e53338c4bba7e2ad3/lib/jwt/jwk/key_base.rb#L11) and just saving it after removing `key_generator` (note that `kid` is actually one of the custom parameters, and can be saved among them :) ).
When the JWK is later exported, any keys in the saved `options` are merged into the result.
One step up from this would be a method to access and manipulate these `options` (maybe even as usability-friendly `my_jwk['use']`).

I would be happy to hear some feedback on this idea and whether this endeavor and whether it may be beneficial to the project.

anakinj:
Also the key accessor feels like a natural way to access JWK attributes. 👍 for that.

anakinj:
Good points.

Are you interested in taking a shot on implementing something like this? Also open for pairing if you're into that :) 

[...]

PR #520, opened 2022-10-07
JWK Common Parameters

bellebaum:
This PR Draft addresses #518 

It adds a new member `common_parameters` to JWKs, representing values in the [IANA JSON Web Key Parameters Registry](https://www.iana.org/assignments/jose/jose.xhtml#web-key-parameters) which are not specific to any particular value of `kty`, as well as any custom values to be associated with a JWK.

There is currently no validation according to the semantics of any of these parameters.
Common Parameters (as they were called in RFC 7517) are automatically imported and exported, and can be explicitly specified for `initialize` using the `common_parameters` option.
During import, specified common parameter keys are converted to symbols.

This PR is mostly API compatible to `main`, with one exception:

When trying to import a JWK using the wrong class (as indicated by the `kty` value), the lib now raises a `JWT::JWKError` before passing the invalid key to OpenSSL and having an error thrown there.
The same happens for unspecified `kty`, which actually makes the import candidate a non-JWK according to [RFC 7517](https://www.rfc-editor.org/rfc/rfc7517.html#section-4.1):

> The "kty" value is a case-sensitive string.  This member MUST be present in a JWK.

Never ending list of things to do:

- [x] Add proper documentation
- [x] Decide whether to reject registered `kty` specific parameters in the `common_parameters` member
- [x] Provide an API-compatible way for future validation of parameters (e.g. `use` would be a good candidate to enforce for the `key_finder`)


lib/jwt/jwk/key_base.rb
```diff
@@ -15,14 +15,19 @@ def initialize(options)
           options = { kid: options }
         end
 
-        @kid           = options[:kid]
+        @common_parameters = options[:common_parameters] || {}
```
anakinj, line 18: Can we use some other word than `common`, feels to generic :). I like the word `parameter` because it's what is used in the RFC.

My suggestion would be`jwk_parameters`

lib/jwt/jwk/key_base.rb
```diff
@@ -15,14 +15,19 @@ def initialize(options)
...
+        @common_parameters = @common_parameters.transform_keys(&:to_sym) # Uniform interface
+        @common_parameters[:kid] = options[:kid] if options[:kid] # kid can be specified outside common_parameters
+
         @kid_generator = options[:kid_generator] || ::JWT.configuration.jwk.kid_generator
       end
 
       def kid
-        @kid ||= generate_kid
+        @common_parameters[:kid] ||= generate_kid
```
anakinj, line 26: The attr accessor could be used here

lib/jwt/jwk/rsa.rb
```diff
@@ -27,7 +27,8 @@ def public_key
       end
 
       def export(options = {})
-        exported_hash = members.merge(kid: kid)
+        kid # Make sure a kid is generated
```
anakinj, line 30: Could we avoid calling the kid method just to populate the parameters somehow. Would it be bad to generate the kid on initialization of the class?

lib/jwt/jwk/rsa.rb
```diff
@@ -70,7 +71,14 @@ def import(jwk_data)
           pkey_params = jwk_attributes(jwk_data, *RSA_KEY_ELEMENTS) do |value|
             decode_open_ssl_bn(value)
           end
-          new(rsa_pkey(pkey_params), kid: jwk_attributes(jwk_data, :kid)[:kid])
+          parameters = jwk_data.transform_keys(&:to_sym)
+          jwk_kty = parameters.delete(:kty) # Will be re-added upon export
+          RSA_KEY_ELEMENTS.each { |e| parameters.delete e }
```
anakinj, line 76:
```suggestion
          RSA_KEY_ELEMENTS.each { |e| parameters.delete(e) }
```

lib/jwt/jwk/rsa.rb
```diff
@@ -83,15 +91,9 @@ def jwk_attributes(jwk_data, *attributes)
...
-        def rsa_pkey(rsa_parameters)
-          raise JWT::JWKError, 'Key format is invalid for RSA' unless rsa_parameters[:n] && rsa_parameters[:e]
-
-          create_rsa_key(rsa_parameters)
-        end
-
         if ::JWT.openssl_3?
           ASN1_SEQUENCE = %i[n e d p q dp dq qi].freeze
-          def create_rsa_key(rsa_parameters)
+          def rsa_pkey(rsa_parameters)
```
anakinj, line 96: Why not `create_*`?

lib/jwt/jwk/ec.rb
```diff
@@ -115,12 +122,6 @@ def to_openssl_curve(crv)
 
         private
 
-        def jwk_attrs(jwk_data, attrs)
```
anakinj, line 118: This is great. Getting rid of stuff required by pre 2.5 Rubies.

lib/jwt/jwk/key_base.rb
```diff
@@ -15,14 +15,19 @@ def initialize(options)
...
       end
 
+      attr_accessor :common_parameters
```
anakinj, line 29: This is public just because of the tests? I guess if we go with the suggested key accessor approach we can make this one private?

anakinj (commented):
Looks great so far. Left some comments, please do not take them as the truth just comments and my opinions and my opinions change daily :)

lib/jwt/jwk/rsa.rb
bellebaum: I was thinking about this as well, but thought at the time that there may be applications which would rather not spend the time computing a kid.

Thinking about it again now, the only flow in which a kid wouldn't be generated anyway is when importing a JWK somehow (probably usually via the `jwks` JWT decoding parameter) and later on only accessing the key pair (i.e. never exporting it). But those cases will most likely always use JWKSs from some sort of public interface, which _should_ already have `kid`s. And if some application really was this limited on resources, they could always replace the `kid` generating algorithm by some dummy alg. (And they probably wouldn't use Ruby for time-critical applications.)

So yes, populating it upon initialization seems like a good idea after all :)

lib/jwt/jwk/rsa.rb
bellebaum: There was an intermediate function `rsa_pkey` which was only conditionally throwing an error and then calling the other one.
At some point it made sense to move the error checking to `import` instead (this is also consistent with **EC** and **oct**), so `rsa_pkey` simply became an alias for `create_rsa_key`. Having an alias for a private function did not make sense IMO, so I removed the indirection. Why `rsa_pkey` instead of `create_rsa_key`? No reason, except that there is a function `ec_pkey` already, so this seemed more consistent to a reader.

lib/jwt/jwk/key_base.rb
bellebaum: I had scanned the RFC for some name to use and these kinds of parameters were only ever termed `common parameters` or `key parameters that are not algorithm specific`.
I agree that `common_parameters` may be too generic. `jwk_parameters` or `parameters` on the other hand do not seem to be restrictive enough for parameters. E.g. in an RSA key I would intuitively expect to be able to set `n` inside the `jwk_parameters`.

Here are some more ideas, none of which are really great IMO:
- `auxiliary_parameters` or `aux_params`
- `additional_parameters`
- `descriptive_parameters` or `desc_params`

lib/jwt/jwk/rsa.rb
bellebaum: I have now reverted the name change and instead renamed `ec_pkey` to `create_ec_key` as well. This is probably the most readable solution :)

lib/jwt/jwk/key_base.rb
anakinj: This is kinda invalidating the initial idea, but what if we would not have these parameters as "exta" but move the importing into the object itself?

Some pseudo usage example of what I was thinking:
```
jwk_data = { kty:'RSA', kid:'thekid', n: 'xx' e: 'xx' d: 'xx' p: 'xx' q: 'xx' dp: 'xx' dq: 'xx' qi: 'xx', use:'sig' }

jwk = ::JWT::JWK.new(jwk_data) # still alias to create_from

jwk.is_a?(::JWT::JWK::RSA)
jwk.keypair # Would generate the RSA key from the parameters or keypair would be generated on initialization
jwk.export(include_private) == jwk_data 
jwk[:use] == 'sig'
jwk[:d] == 'xx'

jwk = ::JWT:JWK.new(OpenSSL::PKey::RSA.new(2048), { kid: 'thekid', use:'sig' })
jwk.is_a?(::JWT::JWK::RSA)

jwk = ::JWT:JWK.new(OpenSSL::PKey::RSA.new(2048), { kid: 'thekid', use:'sig', n:'xxx'} ) # Would throw an error about mixing keypair and parameters is now allowed
````

Comment (bellebaum) [reply lib/jwt/jwk/key_base.rb:18]: I like this idea :)
This is probably the most intuitive solution available for anyone using JWKs (especially the invocation of `new` with OpenSSL key and parameters).

However, the thing which may bite us later is any additional library specific parameters like `kid_generator`, which are no longer distinguishable from JWK parameters. Hypothetical scenario:

Someone wants to explicitly denote within a JWK the generation algorithm for the `kid` within a given ecosystem. Maybe there are at some point several such algorithms in standardized form and a JWK parameter `kid_generator` is being used to explicitly list the subset of algorithms being used in token headers.

Why would someone do this? No idea :)
But it would become a problem for us, as could any other parameter we might want to specify at this point (e.g. we were talking about validation options at one point).

We could pass two hashes to avoid the naming problem. An options hash and a parameters hash (in this order, to not break existing examples in the README):

```ruby
jwk = ::JWT:JWK.new(OpenSSL::PKey::RSA.new(2048),
                    { kid_generator: ::JWT::JWK::Thumbprint },
                    { kid: 'thekid', use:'sig' })
```
The rules are:
- Both the second and third parameter are optional
- The first parameter alone determines the key type and key, no matter whether it is given as a JWK hash or as an OpenSSL key
- The third parameter, if present, may not contain cryptographically relevant material
- For non-key stuff, the third parameter takes priority. E.g. if I specify `use` in both the first and third parameters, the latter will be used.
- The exception is `kid`, which may be present in either parameter for backwards compatibility. I would still argue that the priority should be `param_1 < param_3 < param_2`, but 2 and 3 may as well be the other way around

anakinj: I would like to see the more used parameters as the second/first parameter and the rest that control behaviour as the last one. Allowing `kid_generator` to be given in both for now. I don't see the kid_generator being part of a official RFC any time soon, you can freely quote me and make fun of me when that happens and we are in trouble:)

The keypair problem could be handled by memoizing the generated key until a jwk parameter changes.

bellebaum: Swapped the parameters and implemented memoization. `kid` can no longer be specified in the `options` hash, where any future parameters for validation will be specified.
`kid_generator` can currently be specified in either the options or params hash.
I am going to merge this into the PR :)

lib/jwt/jwk/ec.rb
```diff
@@ -11,37 +11,44 @@ class EC < KeyBase # rubocop:disable Metrics/ClassLength
...
-        raise ArgumentError, 'keypair must be of type OpenSSL::PKey::EC' unless keypair.is_a?(OpenSSL::PKey::EC)
+        # For backwards compatibility when kid was a String
+        params = { kid: params } if params.is_a?(String)
 
-        @keypair = keypair
+        # Accept OpenSSL key as input
+        keypair = parse_ec_key(keypair) if keypair.is_a?(OpenSSL::PKey::EC)
 
-        super(options)
+        raise ArgumentError, 'keypair must be of type OpenSSL::PKey::EC' unless keypair.is_a?(Hash)
```
anakinj, line 27: I have a few suggestions for the initializer (the suggestion applies for all the different types).
 
- The `keypair` parameter could be renamed to something more describing.
- The given OpenSSL object could be kept to avoid recreation from JWK parameters
- The ArgumentError could tell that the parameter also can be a Hash
- Transforming the keys to symbols is only needed when first parameter is a Hash.
 
one idea would be to handle the different variants of the first parameters in a case-block:

```
def initialize(keypair_or_params, params = nil, options = {})
  params ||= {}

  # For backwards compatibility when kid was a String
  params = { kid: params } if params.is_a?(String)

  key_params = case keypair_or_params
               when OpenSSL::PKey::EC  # Accept OpenSSL key as input
                 @keypair = keypair_or_params # Preserve the object to avoid recreation
                 parse_ec_key(keypair_or_params)
               when Hash
                 keypair_or_params.transform_keys(&:to_sym)
               else
                 raise ArgumentError, 'keypair_or_params must be of type OpenSSL::PKey::EC or Hash with key parameters'
               end

  params = params.transform_keys(&:to_sym)

  check_jwk(keypair, params)

  super(options, key_params.merge(params))
end
```

lib/jwt/jwk/key_base.rb
```diff
@@ -8,28 +8,34 @@ def self.inherited(klass)
...
 
-        if options.is_a?(String) # For backwards compatibility when kid was a String
-          options = { kid: options }
-        end
+        @parameters = params.transform_keys(&:to_sym) # Uniform interface
 
-        @kid           = options[:kid]
-        @kid_generator = options[:kid_generator] || ::JWT.configuration.jwk.kid_generator
+        # For backwards compatibility, kid_generator may be specified in the parameters
+        options[:kid_generator] ||= @parameters.delete(:kid_generator)
```
anakinj, line 17: The `kid_generator` value will not be deleted from `@parameters` if `options` contains the `kid_generator`

lib/jwt/jwk/ec.rb
anakinj: What do you think, would an example for the usage without the OpenSSL object in the readme be valuable?

anakinj (commented):
I added one big comment about the JWK initializer, automatically becomes a little messy when backwards compatibility needs to be preserved.

The tests are also failing for older Rubies because of the `Hash#except` method is missing, the exclusion of the private parts needs to be done manually I guess.

Im really liking the improvements, super work! and apologies for my nitpicking:)

bellebaum:
I am having a bit of a struggle with RuboCop.
Locally, when I execute `bundle exec rubocop`, the code only passes all checks if I explicitly disable the class length check for `JWT::JWK::RSA` (which is in large parts due to the different versions of `create_rsa_key`), but on GitHub, it seems to complain about me doing this.
The version seems to be `1.31.2` in both cases.

anakinj:
I think the reason is the runtime you are using. RuboCop is executed on Ruby 3.1 on the CI

lib/jwt/jwk/key_base.rb
bellebaum: This was kinda intended, though poorly implemented.
The situation we are worried about there involves someone specifying `kid_generator` in both `options` and `params`.
Think about different kinds of application programmers:

One kind of programmer is using the API in its current form. They will only specify `kid_generator` in the parameters, since currently, there is no third argument for `initialize`.

The second kind of programmer will use the new JWK API additions. The new README says that `kid_generator` needs to be specified in the third argument. And if they happen to look at the code, they will see the words `backwards compatibility`, and the handling right below.

This latter kind of programmer will probably also expect that anything in the `params` argument (which is not kty-related) will stay a part of the JWK. In some rare circumstances, this will include `kid_generator` for some obscure application (Do not worry, I _will_ quote you when that happens ;D). By specifying the `kid_generator` explicitly in the `options`, they have a workaround if this is indeed their expected behavior.

I do not expect any other kinds of programmers to actively make use of the option to specify `kid_generator` twice.
What do you think?

anakinj: Everything makes sense, just leave it like this. It's simple

lib/jwt/jwk/ec.rb
bellebaum: > What do you think, would an example for the usage without the OpenSSL object in the readme be valuable?

Yes. In fact, the section `Importing and exporting JSON Web Keys` should probably demonstrate this instead of converting an OpenSSL key as in the example above.

> The `keypair` parameter could be renamed to something more describing.

I agree, though `keypair_or_params` runs the risk of suggesting that you can specify everything from the first argument in the second, which is not the case (e.g. the first argument needs to be the one describing the `kty`).

In some other places I think I have simply used `key`, which could be an OpenSSL Key or a JSON Web Key.

> The given OpenSSL object could be kept to avoid recreation from JWK parameters

I like this :+1:

bellebaum:
There is one more feature I have been thinking about adding:

When initializing a JWK with another JWK of the same type as `key`, it should make a copy of this key.

This would allow stuff such as

```ruby
jwk1 = JWT::JWK.new(OpenSSL::PKey::RSA.new 2048)
jwk2 = JWT::JWK.new(OpenSSL::PKey::RSA.new 2048)
jwks = { keys: [jwk1, jwk2] } # Note: no explicit export

# Uses key accessor to find key with right `kid` and calls `initialize` and `keypair`
JWT.decode my_token, nil, true, jwks: jwks
```

which is useful if the JWKS is not created as above but fetched from some URL, but I guess even more useful would be some JWKS handling, which is best handled in a future PR :)

anakinj:
I like the suggestion. But maybe not in this PR. 

One tiny thing still related to this PR, a little note in the changelog would be great.

anakinj (approved):

anakinj:
We are happy with this, right? No more changes coming @bellebaum?

bellebaum:
Not in this PR. I think it should be possible to extend this in the future, given that options parameters are now separate from params. So I am ok with the current state being merged unless there are further suggestions

anakinj:
Great. Was thinking we could make some plan for the 3.0 version of this gem to get rid of the older behaviour of the interfaces that clutter the code.

anakinj:
Super work here, highly appreciated. Also the valuable suggestions and feedback on these JWK features is great!

Issue #523, opened 2022-10-11
RSA JWK parsing


bellebaum:
This came up while reworking the JWK stuff for #520 

The function `create_rsa_key` for OpenSSL version 3 looks like this:

https://github.com/jwt/ruby-jwt/blob/b4f9f1d483086054e812dc6ce5655466b6de3858/lib/jwt/jwk/rsa.rb#L93-L106

I like the approach of converting a JWK to DER, then feeding it into OpenSSL, since this is unlikely to change (in fact, wouldn't this work for other versions of OpenSSL, making all the "legacy" implementations unnecessary?)

Anyways, the code has a few problems.

Firstly, any subset of parameters specified in `ASN1_SEQUENCE` could be present in an illformed JWK, and the resulting DER can then probably not be parsed by any SSL library.

But secondly, there is an important missmatch between [RFC 7518, Section 6.3.2](https://www.rfc-editor.org/rfc/rfc7518.html#section-6.3.2)

> The parameter "d" is REQUIRED for RSA private keys.  The others enable optimizations and SHOULD be included by producers of JWKs representing RSA private keys. If the producer includes any of the other private key parameters, then all of the others MUST be present, with the exception of "oth", which MUST only be present when more than two prime factors were used.

and [RFC 3447, Appendix A.1.2](https://www.rfc-editor.org/rfc/rfc3447#appendix-A.1.2), which only permits keys with all parameters present.
This implies that valid JWKs containing exactly the `kty` specific keys `kty`, `n`, `e`, and `d`, are not guaranteed to be parseable this way.

One could try to calculate the missing values, but especially the step of finding `p` and `q` is not trivial at all...

anakinj:
There are some good points here. I think we should extend the test suite for the JWKs for "incomplete" parameters.

Like how does OpenSSL 2 and 3 differ in behaviour. I would not yet be confident using the DER approach as the primary way to generate keys.

Are you planning doing something about this or should we take one challenge at a time and try to finish the JWK importing that is WIP?

bellebaum:
I was trying to do something briefly today, but decided that this will probably take more time than I can spend right now, so I am focussing on existing challenges and leaving this here as a reminder or for someone else to pick up in the meantime :)

anakinj:
I will take a stab on trying to expand the test suite a little. Lets see if there are some differences in behaviours with the implementations.

bellebaum:
[...]

PR #524, opened 2022-10-12
Stricter RSA key generation from jwk parameters

anakinj:
This PR is addressing a parts of #523 and adding test coverage on a few scenarios. 

Moved the methods around a little for easier testing and added parameter validation with the rule:
 - If one of the optimisation params are given ALL needs to be given

As you @bellebaum pointed out the DER generation has some issues when parameters are missing. Did not dig that deep but I have a feeling that there is no way to present a private key in the DER format without these parameters.

Im a little unsure what do, not really eager into starting to calculate the primes etc. I guess it's doable but feels sketchy. 

So I guess the question is:

Is there a way to generate a usable RSA object with OpenSSL 3.0 with only the modulus and exponents (n,e,d)? as the JWK spec allows private keys to be presented with only these values.

### DIFF
--- a/lib/jwt/jwk/rsa.rb
+++ b/lib/jwt/jwk/rsa.rb
@@ -25,7 +25,7 @@ module JWT
       end
 
       def keypair
-        @keypair ||= create_rsa_key(jwk_attributes(*(RSA_KEY_ELEMENTS - [:kty])))
+        @keypair ||= self.class.create_rsa_key(jwk_attributes(*(RSA_KEY_ELEMENTS - [:kty])))
       end
 
       def private?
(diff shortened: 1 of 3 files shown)


bellebaum:
I have asked about this over at the Ruby OpenSSL gem: https://github.com/ruby/openssl/issues/551
Seems like this is a TODO for them, so we should eventually see a way.

anakinj:
Thanks @bellebaum for raising the question. Im going to return to this one to have it working with the adjustments from #520 at some point.

anakinj:
This is the best we can to with the interfaces provided by the OpenSSL gem. The behaviour is going to be different depending on the version of openssl that is in use.

anakinj:
Im going to merge this soonish.

Working on https://github.com/ruby/openssl/pull/555 to move all this heavy lifting into the openssl gem. Hopefully it will be great some day :)

PR #525, opened 2022-10-17
Add a JWKS implementation

bellebaum:
This PR Draft implements JWK Sets (Part of [RFC 7517](https://www.rfc-editor.org/rfc/rfc7517.html)) for easier handling of sets.

The goal was to make the handling of JWKS easier. E.g.

```ruby
json = Net::HTTP.get(my_friends_jwks_uri)
jwks = JWT::JWK::Set.new(JSON.parse(json))
jwks.filter! { |key| key[:use] == 'sig' } # Signing Keys only
algorithms = jwks.map { |key| key[:alg] } # & algs_allowed_by_local_policy
JWT.decode some_token, nil, true, algorithms: algorithms, jwks: jwks

my_jwks = JWT::JWK::Set.new # Empty JWKS
my_jwks << JWT::JWK.new(OpenSSL::PKey::RSA.new(2048), use: 'sig')
my_jwks << OpenSSL::PKey::RSA.new(2048)
my_jwks.merge(jwks)
hash = my_jwks.export
```

The draft allows for basic manipulation of the set:

- adding JWKs
- merging JWKSs
- filtering JWKs inside a JWKS
- `export`ing to Hash

Where possible, I tried to use function names and semantics common in the standard library, to not subvert user expectations.

There are currently no tests or documentation, but I would like to get some early feedback to incorporate :)
The selection of methods on JWKSs is currently based on stuff I find helpful in my projects, so feel free to request any changes.

### DIFF
--- a/lib/jwt/jwk/key_finder.rb
+++ b/lib/jwt/jwk/key_finder.rb
@@ -5,8 +5,12 @@ module JWT
     class KeyFinder
       def initialize(options)
         jwks_or_loader = options[:jwks]
-        @jwks          = jwks_or_loader if jwks_or_loader.is_a?(Hash)
-        @jwk_loader    = jwks_or_loader if jwks_or_loader.respond_to?(:call)
+
+        @jwks_loader = if jwks_or_loader.respond_to?(:call)
+          jwks_or_loader
+        else
+          ->(_options) { jwks_or_loader }
+        end
       end
 
       def key_for(kid)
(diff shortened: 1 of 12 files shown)


lib/jwt/jwk/key_finder.rb
```diff
@@ -5,57 +5,37 @@ module JWK
     class KeyFinder
       def initialize(options)
         jwks_or_loader = options[:jwks]
-        @jwks          = jwks_or_loader if jwks_or_loader.is_a?(Hash)
-        @jwk_loader    = jwks_or_loader if jwks_or_loader.respond_to?(:call)
+
+        @jwks_loader = if jwks_or_loader.respond_to?(:call)
+          jwks_or_loader
+        else
+          ->(_options) { jwks_or_loader }
```
anakinj, line 12: Using only procs is a great idea to avoid complexity later.

lib/jwt/jwk/set.rb
```diff
@@ -0,0 +1,80 @@
...
+                else
+                  raise ArgumentError, 'Can only create new JWKS from Hash, Array and JWK'
+        end
+      end
+
+      def export(options = {})
+        { keys: @keys.map { |k| k.export(options) } }
+      end
+
+      def each(&block)
```
anakinj, line 32: `Forwardable` could be used to delegate methods to the underlying Array.

anakinj:
This is a great idea. Left a few random comments

bellebaum:
The `uniq!` bug seems strange.
It sometimes works for me using the `openssl` gemfile, and I was hoping that this was some setup-issue, but apparently it is not.
[The documentation](https://ruby-doc.org/core-3.1.2/Array.html#method-i-uniq-21) seems to indicate that `uniq!` on Arrays filters duplicates as indicated by comparison using `.eql?`. Comparing the two elements manually using this method shows that they are indeed equal, yet `uniq!` does not find any duplicates. I think I am missing something obvious here.

anakinj:
The failing test is probably because https://github.com/ruby/openssl/pull/538 has been shipped. Need to make the tests somehow pass on all versions. I'll take a look at this at some point in the near future, on a quick glimpse it looks great.

anakinj (approved):

anakinj:
@bellebaum I fixed the CI for `ruby-head` could you be so kind and rebase/merge main into your branch?

Also a changelog entry would be great in regards to this feature addition.

anakinj:
Great stuff. Big thank you for putting effort into this!

Issue #394, opened 2020-12-21
Support for OKP JWK keys


anakinj:
We could have support for the representing Ed25519 keys as JWK. To widen the support for the different algorithms.

This was already started here: #388, but the JWK classes need some rework to logically fit the "signing-only" algorithm. This work is ongoing and after that finishes the initial implementation for the OKP keytype will be trivial.

https://tools.ietf.org/html/rfc8037

anakinj:
Released in 2.7.0. Let's return to this when ruby/openssl has better parameter support for this key type.

PR #540, opened 2022-12-28
JWK OKP Ed25519 support using rbnacl

anakinj:
- Support OPK JWKs, only Ed25519 for now
 - Requires the rbnacl gem
 - Prepare to deprecate the ::JWT::JWK::[RSA/EC/HMAC]#keypair method in favor of methods describing the use

Related to #394

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 **Features:**
 
+- Support OKP (Ed25519) keys for JWKs [#540](https://github.com/jwt/ruby-jwt/pull/540) ([@anakinj](https://github.com/anakinj)).
 - Your contribution here
 
 **Fixes and enhancements:**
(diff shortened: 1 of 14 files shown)


cseeman:
Hey there @anakinj , any idea when this might merge? Would love to get that Ed25519 support in ruby-jwt so we can use that stronger cipher.

timdawborn:
+1. Would love to see Ed25519 support in this library.

anakinj:
@cseeman and @timdawborn. Im thinking of merging this and cutting a new release. Could you be so kind and provide some feedback on does it work "for real" everything so far is tested "in theory".

Also when openssl ruby supports these types better i think we could drop the rbnacl dependency.

cseeman:
Hey @anakinj just wanted to make sure that you got some feedback on this. It has been working in a prod system for a while with the new Ed25519 support, using JWK keys that are making using of this cipher. Thank you again for getting this merged in.

anakinj:
So appreciated that you took time to come back for the feedback. Great to hear it's valuable.

PR #543, opened 2023-01-23
`allow_nil_kid` option for the JWK KeyFinder

bellebaum:
When a JWT does not specify a `kid` explicitly in its header, its key may still have to be fetched from a JWKS.
This may happen e.g. when there is only one key in the JWKS and thus there is no need to specify an ID.

This PR adds a decoding option `:allow_nil_kid`, which, when used in conjunction with the `:jwks` option, prevents an error from being raised when the JWT does not specify the `kid`.
In this case, the first key in the provided JWKS is used as the decryption key.

To prevent application jwks-handlers from being overwhelmed by a sudden nil in the options hash, the functionality is disabled by default. (To be fair though, there are no type checks on `kid` and this is all before any verification has happened, so applications _should_ have implemented the necessary checks anyways. So there is an argument that one could just make this the default. :D)

### DIFF
--- a/lib/jwt/jwk/key_finder.rb
+++ b/lib/jwt/jwk/key_finder.rb
@@ -4,6 +4,7 @@ module JWT
   module JWK
     class KeyFinder
       def initialize(options)
+        @allow_nil_kid = options[:allow_nil_kid]
         jwks_or_loader = options[:jwks]
 
         @jwks_loader = if jwks_or_loader.respond_to?(:call)
@@ -14,7 +15,7 @@ module JWT
       end
 
       def key_for(kid)
-        raise ::JWT::DecodeError, 'No key id (kid) found from token headers' unless kid
+        raise ::JWT::DecodeError, 'No key id (kid) found from token headers' unless kid || @allow_nil_kid
 
         jwk = resolve_key(kid)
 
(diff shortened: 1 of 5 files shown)


anakinj:
Makes sense, gets a little hacky like this when we need to keep backwards compatibility. Would you still be kind to add a changelog entry into the `CHANGELOG.md` file?

bellebaum:
I went ahead and added a simple type check on the `kid` value in the keyfinder.

I think there are few applications which handle the case where `kid` is not a string but fail if it is `nil`. So making this the default _should_ not break applications that do not need fixing anyway. But there is of course the remote chance that someone may have taken a look at this gem's code, saw the `nil`-check and decided to take that for granted.

Personally, I think that if such people exist, they will also be able to fix their code easily and would be comfortable taking a chance here, but strictly speaking it is a backwards-incompatible change and you may also simplify this with the next major version.

Your choice :)

lib/jwt/jwk/key_finder.rb
```diff
@@ -14,7 +15,8 @@ def initialize(options)
       end
 
       def key_for(kid)
-        raise ::JWT::DecodeError, 'No key id (kid) found from token headers' unless kid
+        raise ::JWT::DecodeError, 'No key id (kid) found from token headers' unless kid || @allow_nil_kid
+        raise ::JWT::DecodeError, 'Invalid type for kid header parameter' unless kid.nil? || kid.is_a?(String)
```
anakinj, line 19: I think `unless kid.is_a?(String)` is enough. nil is also a "type"

anakinj, line 19: I think checking just if the kid is a String is enough, `nil.is_a?(String) == false`
```suggestion
        raise ::JWT::DecodeError, 'Invalid type for kid header parameter' unless kid.is_a?(String)
```

anakinj:
I think it's great. Just added very small suggestion:)

lib/jwt/jwk/key_finder.rb
bellebaum: For my Ruby version (3.0.5), I get the following:

```ruby
p nil.is_a(String)
```

```
-:1:in `<main>': undefined method `is_a' for nil:NilClass (NoMethodError)
```

Even then, the whole point of this PR is to make this series of checks evaluate to `true` for the `unless` clauses in the presence of `nil`, so errors are not automatically raised if the application can handle them :)

anakinj: `is_a?` should be pretty basic functionality since the early days.

bellebaum: Nvm, I have missed a `?` :sweat_smile:
Still, if it returns `false`, `unless` will not prevent the error from being raised.

anakinj: Youre right it needs to pass that line if it's nil. `unless` is so hard for the brain...

anakinj (approved):

anakinj:
Thank you for your contribution.

PR #544, opened 2023-01-24
Automatic JWK enrichment

bellebaum:
This PR will allow to derive most standardized common parameters from already specified parameters.
This is a follow-up to a few points which could not be addressed in #520, but were brought up by me in #518.

# Overview

Using all of its features, given an X.509 certificate, you can acquire a JWK like this:

```ruby
::JWT::JWK.new(certificate)
```

This will extract the public key, construct the JWK, and try to set the following parameters:

- `kid` as previously implemented
- `x5c` is a representation of the certificate
- `x5t` and `x5t#S256` are wide-spread certificate thumbprints
- `key_ops` and `use` state allowed key usages and are derived from the X.509 `keyUsage` extension, if present
- `alg` will be set to a sensible default depending on `use` and `kty`

Of course, any parameters specified in the above invocation will still take precedence.

There is even support for the `x5u` parameter, to fetch a remote certificate chain:

```ruby
x5u_fetcher = ->(uri) {
    # Fetch uri, return as Array of OpenSSL::X509::Certificate
}
::JWT::JWK.new(jwks_hash, nil, enrich_key: true, x5u_handler: x5u_fetcher)
```

# TODO

- [ ] Proper certificate validation
- [ ] Check if code from `x5c_key_finder` can be reused
- [ ] new README examples
- [ ] Update changelog

### DIFF
--- a/lib/jwt/jwk/key_base.rb
+++ b/lib/jwt/jwk/key_base.rb
@@ -19,6 +19,8 @@ module JWT
         # Make sure the key has a kid
         kid_generator = options[:kid_generator] || ::JWT.configuration.jwk.kid_generator
         self[:kid] ||= kid_generator.new(self).generate
+
+        enrich_key(options) if options[:enrich_key]
       end
 
       def kid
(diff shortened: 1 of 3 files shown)


anakinj:
I realize this is a WIP but I got very curious so will drop a few comments.

lib/jwt/jwk/key_base.rb
```diff
@@ -50,6 +52,83 @@ def <=>(other)
       private
 
       attr_reader :parameters
+
+      KEY_USAGES_SIG  = ['Digital Signature', 'Non Repudiation', 'Content Commitment', 'Key Cert Sign', 'CRL Sign'].freeze
```
anakinj, line 56: Im wondering about where this certificate specific stuff should live. Now the logic baked into each and every JWK object created. 

Most of these methods are just setting some values under certain keys? Could it just be a separate mechanism (method/extension) that would build the params passed into the key or add the params to an existing key?

anakinj:
@bellebaum are you planning continuing on this one at some point? I'm a bit in a summer cleaning mode and tempted to close this as stale :)

bellebaum:
No, unfortunately this is stale for the moment.
We were going to use this in an OAuth Server, but that one has not seen much development as of late, and it might take a bit until someone picks it up for the next research project. So by extension, I currently cannot work on this :/

anakinj:
No problem. I'll close this for now.

PR #545, opened 2023-01-27
Fix traversing not matching header algos

mpospelov:
Hello ruby-jwt maintainers! Thank you for these amazing tool. 🙏

Recently in our company we started to thinking of how we could rotate keys with 0 downtime. 

I found that `keyfinder` can actually return array of keys and `ruby-jwt` will iterate through these keys until find a matching one
https://github.com/jwt/ruby-jwt/blob/8963312352f1698091640f3201e930c201ea933e/lib/jwt/decode.rb#L47

**BUT** the problem is that for each key we traverse all allowed algoes in decoder. 
Our provider code supports multiple consumers which encodes payload with different `algo`es, so the current code can lead to the case when we have matching public key returned as second element in keyfinder but we even don't try to decode the payload with it as we fail with attempt to decode with first key but with incorrect algo.

You can find these case in specs and implementation to fix this issue in this PR.


lib/jwt/decode.rb
```diff
@@ -49,10 +49,12 @@ def verify_signature
       raise(JWT::VerificationError, 'Signature verification failed')
     end
 
-    def verify_algo
+    def verify_and_set_algos
       raise(JWT::IncorrectAlgorithm, 'An algorithm must be specified') if allowed_algorithms.empty?
       raise(JWT::IncorrectAlgorithm, 'Token is missing alg header') unless alg_in_header
-      raise(JWT::IncorrectAlgorithm, 'Expected a different algorithm') unless valid_alg_in_header?
+
+      @algos = find_valid_algos_in_header
```
anakinj, line 56: Setting the `@algos` in the middle of everything feels a little wrong.

We are interested in the intersection between the allowed_algorithms and what is specified in the token header, right?

Could there be a method (maybe a memoizing one) that would be used in the `verify_algos` and `verify_signature_for?`?
for example
```
def allowed_and_valid_algorithms
  @allowed_and_valid_algorithms ||= allowed_algorithms.select { |alg| alg.valid_alg?(alg_in_header) }
end
```

anakinj (changes requested):

anakinj:
Hi, I think I get the problem and I don't see a problem filtering the list based on the user provided algorithms. I added a comment on the way the algorithms to be used are resolved.

mpospelov:
Hey @anakinj thank you for your kind review! I addressed rubocop fails and implemented the `#allowed_and_valid_algorithms` as you suggested

mpospelov:
also `CHANGELOG.md` entry is added 😊

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@
 
 **Fixes and enhancements:**
 
-- Your contribution here
+- Fix issue with multiple keys returned by keyfinder and multiple allowed algorithms[#545](https://github.com/jwt/ruby-jwt/pull/545) ([@mpospelov](https://github.com/mpospelov))
```
anakinj, line 14: Nitpick: Lets keep the placeholder for the next person.
```suggestion
- Your contribution here

```

anakinj (changes requested):

anakinj (approved):

anakinj:
Great improvement. Thank you for the effort.

mpospelov:
Hey @anakinj! Thank you for all the support and for reviewing my PR 😊 One last thing is don't you mind releasing a new version of `gem`? It would help us a lot 🙏

anakinj:
New version out. Have fun with it :)

PR #552, opened 2023-02-03
Fix typo in README

tatsuyafw:
`allow_nil_kid` option was introduced in #543, but the option name was wrong in README.

This change fixes the typo.


anakinj:
Super, thanks for the quick reaction.

tatsuyafw:
Thank you for reviewing!

PR #555, opened 2023-03-04
fix signature has expired error if payload is a string

GobinathAL:
currently it's possible to encode a string payload and decode it but decoding a string containing "exp" throws signature has expired payload which could be misleading as the payload has to be a hash inorder for this to work properly.

Possible before this PR:
```ruby
payload = "Hello"
secret = "secret"
token = JWT.encode payload, secret
#"eyJhbGciOiJIUzI1NiJ9.IkhlbGxvIg.gR1jkofv3fsS7g6lgRR5pBEuePbpWTuUEQ847GKQ7mc"
JWT.decode token, secret
#["Hello", {"alg"=>"HS256"}]
```

```ruby
payload = "beautyexperts" #contains exp
secret = "secret"
token = JWT.encode payload, secret
#"eyJhbGciOiJIUzI1NiJ9.ImJlYXV0eWV4cGVydHMi.7zl7sGM_oNoo1nCUH22chd0Ofi67n9GEfN0DS1_lR7w"
JWT.decode token, secret
#.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/jwt-2.7.0/lib/jwt/verify.rb:42:in `verify_expiration': Signature has expired (JWT::ExpiredSignature)
```

Reason for making this PR is has_key will atleast throw undefined method error so that the error won't go unnoticed in production if the payload is a string by mistake.

### DIFF
--- a/lib/jwt/verify.rb
+++ b/lib/jwt/verify.rb
@@ -38,12 +38,12 @@ module JWT
     end
 
     def verify_expiration
-      return unless @payload.has_key?('exp')
+      return unless @payload.key?('exp')
       raise(JWT::ExpiredSignature, 'Signature has expired') if @payload['exp'].to_i <= (Time.now.to_i - exp_leeway)
     end
 
     def verify_iat
-      return unless @payload.has_key?('iat')
+      return unless @payload.key?('iat')
 
       iat = @payload['iat']
       raise(JWT::InvalidIatError, 'Invalid iat') if !iat.is_a?(Numeric) || iat.to_f > Time.now.to_f
@@ -77,7 +77,7 @@ module JWT
     end
 
     def verify_not_before
-      return unless @payload.has_key?('nbf')
+      return unless @payload.key?('nbf')
       raise(JWT::ImmatureSignature, 'Signature nbf has not been reached') if @payload['nbf'].to_i > (Time.now.to_i + nbf_leeway)
     end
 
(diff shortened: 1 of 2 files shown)


anakinj (commented):
Im wondering if this is too big of a change for a minor version. This will require the `@payload` to be a Hash or an object responding to `has_key?`. I would also argue that it's totally fine to pass a string as the payload. The RFC states `JWTs encode claims to be transmitted as a JSON [RFC7159] object that is used as the payload of a JSON Web Signature`

GobinathAL:
> I would also argue that it's totally fine to pass a string as the payload

@anakinj 
If we're allowing String as payload, are we giving users the option to pass expiry time? Either string shouldn't be allowed or i would suggest that it is ok if we don't check for expiry if the payload is string. What are your thoughts?

lib/jwt/verify.rb
```diff
@@ -38,12 +38,12 @@ def verify_aud
     end
 
     def verify_expiration
-      return unless @payload.include?('exp')
+      return unless @payload.key?('exp')
```
anakinj, line 41: What if there would be a method on this class, something like 

```
def contains_key?(payload, key)
  payload.respond_to?(:key?) && payload.key?(key)
end
```

Then this row would be
```
return unless contains_key?(@payload, 'exp')
```

anakinj:
I think the proposed changes are great but we need to support the case when the passed payload is not responding to the `#key?` method. 

To my understanding passing a string as the payload is not prohibited. What comes to the expiry, [the ´exp´ claim (header) is optional for a JWT token](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.4).

Added a little suggestion on how we could handle such cases.

anakinj (approved):

anakinj (changes requested):
Totally forgot to ask if there by any chance would be possible to have some kind of test to validate this change?

anakinj (approved):

anakinj:
Great work and thanks for you patience

Issue #558, opened 2023-04-17
Bug: undefined method `casecmp' for nil:NilClass when invalid alg


nataliastanko:
Getting `#<NoMethodError: undefined method casecmp for nil:NilClass>`  when providing invalid algorithm to the `JWT.decode`:

`JWT.decode(token, 'secret', true, algorithm: 'invalid-HS256')`

Affecting the **jwt** gem since the https://github.com/jwt/ruby-jwt/pull/512

PR #559, opened 2023-04-17
Handle invalid algorithm when decoding JWT

nataliastanko:
Fixes https://github.com/jwt/ruby-jwt/issues/558

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 **Fixes and enhancements:**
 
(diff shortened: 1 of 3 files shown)


anakinj:
Looks great, thank you for your effort so far. The issue with the CI seems to be unrelated. Need to take a look at that.

anakinj:
The failing test should be fixed in main now. Would you @nataliastanko be so kind and rebase your branch to verify everything is great?

anakinj (approved):

anakinj:
Thank you for the fix!

Issue #309, opened 2019-05-06
Support for Unencoded Payload Option (RFC 7797)


idigber:
Is there support for JWS Unencoded Payload Option (RFC 7797) for signature calculation using this library? More details can be found here: https://tools.ietf.org/html/rfc7797

excpt:
@idigber This feature is not supported yet. I will keep this open and mark it as a feature request.

wout:
Any progress on this one?

WriterZephos:
We really need this feature. Please add it as soon as you can. Thank you!

WriterZephos:
Okay I got something up: https://github.com/jwt/ruby-jwt/pull/561

@idigber @excpt @wout

abiagini:
Hi guys, this feature is intended to be reviewed?

anakinj:
Support for dealing with unencoded payload was added to the `JWT::EncodedToken` class introduced in version 3.x

PR #561, opened 2023-05-05
Add support for RFC7797 unencoded payloads

WriterZephos:
This adds support for RFC7797 and addresses https://github.com/jwt/ruby-jwt/issues/309

Here is the specification: https://www.rfc-editor.org/rfc/rfc7797

Happy to receive feedback and would love to get this released soon as I need it for my current project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -680,6 +680,72 @@ jwk_hash = jwk.export
 thumbprint_as_the_kid = jwk_hash[:kid]
 ```
 
+### Unencoded and Detached Payloads
+
+#### Unencoded Payloads
+
+To generate a JWT with an unencoded payload, you may use the `b64` header set to false as described by RFC 7797. When you do this, the `crit` header will be added if it doesn't already exist, and the `b64` value will be appended to it.
+
+```ruby
+private_key = RbNaCl::Signatures::Ed25519::SigningKey.new('abcdefghijklmnopqrstuvwxyzABCDEF')
+public_key = private_key.verify_key
+token = JWT.encode payload, private_key, 'ED25519', { b64: false }
+
+# eyJiNjQiOmZhbHNlLCJhbGciOiJFRDI1NTE5IiwiY3JpdCI6WyJiNjQiXX0.{\"data\":\"test\"}.RL6jDz7h_fbQQds1x_ABOVE_dp646ZIbzvBB_DlixrTTMAiG7k0q4wH8dpcQ7KUeGgqI0tqj7B4JG_jTwM6fCg
+puts token
+
+decoded_token = JWT.decode token, public_key, true, { algorithm: 'ED25519' }
+# Array
+# [
+#  {"data"=>"test"}, # payload
+#  {"b64"=>false, "alg"=>"ED25519", "crit"=>["b64"]} # header
+# ]
+```
+
+It is extremely important that one take great care when using unencoded payloads, as the payload must be url safe if it is intended to be transmitted, etc. Also, because `.` is used to delineate between JWT segments, the payload must not have any `.` characters. If the paylod contains `.` then an `InvalidUnencodedPayload` error is raised.
+
+For the above reasons, detached payloads are often used in combination with unencoded payloads.
+
+#### Detached Payloads
+
+To generate a JWT with a detached payload, you must call `encode_detached` instead of `encode`. Then, when decoding and verifying the token, you must pass the `payload` option with the value of the detached payload before encoding.
+
+```ruby
+private_key = RbNaCl::Signatures::Ed25519::SigningKey.new('abcdefghijklmnopqrstuvwxyzABCDEF')
+public_key = private_key.verify_key
+token = JWT.encode_detached payload, private_key, 'ED25519'
+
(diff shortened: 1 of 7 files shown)


lib/jwt.rb
```diff
@@ -22,7 +22,16 @@ def encode(payload, key, algorithm = 'HS256', header_fields = {})
     Encode.new(payload: payload,
                key: key,
                algorithm: algorithm,
-               headers: header_fields).segments
+               headers: header_fields,
+               detached: false).segments
```
abiagini, line 26: Maybe we can set detached as default false

lib/jwt/encode.rb
```diff
@@ -15,11 +15,24 @@ def initialize(options)
       @algorithm        = resolve_algorithm(options[:algorithm])
       @headers          = options[:headers].transform_keys(&:to_s)
       @headers[ALG_KEY] = @algorithm.alg
+      @detached         = options[:detached]
```
abiagini, line 18: Here we can set as default false
```
@detached = options[:detached] || false
```

lib/jwt/decode.rb
```diff
@@ -152,15 +153,35 @@ def header
...
-      @segments.first(2).join('.')
+      [@segments[0], encoded_payload].join('.')
     end
 
-    def parse_and_decode(segment)
-      JWT::JSON.parse(::JWT::Base64.url_decode(segment))
+    def parse_and_decode(segment, decode: true)
+      if decode
+        JWT::JSON.parse(::JWT::Base64.url_decode(segment))
+      else
```
abiagini, line 182: We can avoid else statement with a guard clause.

abiagini (commented):
In general LGTM, I made some minor comments.

abiagini:
Hi guys, this feature is quite common, will be good to have this supported.

WriterZephos:
> Hi guys, this feature is quite common, will be good to have this supported.

Thanks for you suggestions. I will try to find time to push up some updates soon.

lib/jwt/decode.rb
```diff
@@ -152,15 +153,35 @@ def header
...
+    end
+
+    def encoded_detached_payload
+      payload ||= ::JWT::Base64.url_encode(JWT::JSON.generate(@detached_payload)) if decode_payload?
+      payload ||= @detached_payload.to_json
+      payload
+    end
+
+    def decode_payload?
+      header['b64'].nil? || !!header['b64']
```
anakinj, line 172: I wonder if this is a bit problematic to change behavior because of user-input. Would it be too much of an inconvenience to have the caller decide to decode or not?

lib/jwt/decode.rb
```diff
@@ -152,15 +153,35 @@ def header
...
     end
 
     def signing_input
-      @segments.first(2).join('.')
+      [@segments[0], encoded_payload].join('.')
     end
 
-    def parse_and_decode(segment)
-      JWT::JSON.parse(::JWT::Base64.url_decode(segment))
+    def parse_and_decode(segment, decode: true)
```
anakinj, line 179: Would it be possible to avoid the parameter if the caller would just call parse and decode in separate?

lib/jwt/decode.rb
```diff
@@ -152,15 +153,35 @@ def header
     end
 
     def payload
-      @payload ||= parse_and_decode @segments[1]
+      @payload ||= parse_and_decode(encoded_payload, decode: decode_payload?)
+    end
+
+    def encoded_payload
+      payload = encoded_detached_payload if !@detached_payload.nil? && @segments[1].empty?
```
anakinj, line 160: I think it would make sense to raise an error if a payload is given and the token also contains a payload, instead of just ignoring the given payload.

lib/jwt/encode.rb
```diff
@@ -51,7 +64,21 @@ def encode_header
     end
 
     def encode_payload
-      encode_data(@payload)
+      # if b64 header is present and false, do not encode payload as per RFC7797 proposed standard
```
anakinj, line 67: Would it make sense to not rely on header values but instead have an explicit parameter that controls the behavior of both header and payload encoding?

README.md
```diff
@@ -680,6 +680,72 @@ jwk_hash = jwk.export
...
+#  {"alg"=>"ED25519"} # header
+# ]
+```
+
+#### Combining Unencoded and Detached Payload Support
+
+Unencoded and detached payloads are often used hand in hand, such as in proof signatures of Verifiable Credentials. You may use the `b64` header in combination with `encode_detached` to combine both features.
+
+```ruby
+private_key = RbNaCl::Signatures::Ed25519::SigningKey.new('abcdefghijklmnopqrstuvwxyzABCDEF')
```
anakinj, line 734: The example could be a more simple one than one that required the rbnacl dependency.

anakinj (changes requested):
Finally got time to dig a bit into this. I think the direction is nice. A few comments related to maybe being more explicit on how to behave when encoding and decoding.

anakinj:
Some autumn cleanup happening. So going to close this one as stale.

PR #563, opened 2023-06-08
Do not raise error when verifying bad HMAC signature

hieuk09:
Currently, if user inputs token with incorrect signature (example: a part of signature is missing because it was truncated), an error is raised. It can be reproduced using the code below:

```ruby
data = 'a string to be encoded'
key = 'a secret'
token = JWT.encode(data, key, 'HS256')
new_token = token[0..-2]
JWT.decode(new_token, key, true, algorithm: 'HS256') # raise error Provided authenticator was 31 bytes (Expected 32) (RbNaCl::LengthError)
```

It would be great if `Signature verification failed (JWT::VerificationError)` is raised in this case

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 **Fixes and enhancements:**
 
 - Handle invalid algorithm when decoding JWT [#559](https://github.com/jwt/ruby-jwt/pull/559) - [@nataliastanko](https://github.com/nataliastanko)
+- Do not raise error when verifying bad HMAC signature [#563](https://github.com/jwt/ruby-jwt/pull/563) - [@hieuk09](https://github.com/hieuk09)
 - Your contribution here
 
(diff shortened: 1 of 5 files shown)


anakinj:
The change looks great.

But the plan for the next major version is to drop the RbNaCl for HMAC signatures. Is there any reason for you using it or is it just in use because of the behaviour of the gem currently replaces the hmac algorithm handling with RbNaCl if the rbnacl gem is available?

hieuk09:
>  because of the behaviour of the gem currently replaces the hmac algorithm handling with RbNaCl if the rbnacl gem is available?

Yes, this is the reason

hieuk09:
Great, thank you. I'll create a changelog entry for it.

anakinj (approved):

anakinj:
Thank you for the effort of fixing this

PR #569, opened 2023-10-13
Fix key base equality and spaceship operators

magneland:
I ran into this issue when trying to use `Rails.cache.fetch`.

Specifically there is a [fallback condition](https://github.com/rails/rails/blob/6-1-stable/activesupport/lib/active_support/cache.rb#L837) that tried to compare:

`@value == true` (where `@value` was a JWK)

and this gem raised an error:

```
undefined method `[]' for true:TrueClass
```

I haven't run into the spaceship operator issue, but I figured it is similar enough to fix at the same time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 **Fixes and enhancements:**
 
 - Fix signature has expired error if payload is a string [#555](https://github.com/jwt/ruby-jwt/pull/555) - [@GobinathAL](https://github.com/GobinathAL).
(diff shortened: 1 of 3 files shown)


magneland:
FYI I followed some of the recommendations here: https://shopify.engineering/implementing-equality-in-ruby

anakinj (approved):

anakinj:
Thank you for taking time fixing this. 

Would you have the time to add a changelog entry for this into [the changelog file](https://github.com/jwt/ruby-jwt/blob/main/CHANGELOG.md). I had forgotten to update it after the last release, fixed now on main so if you rebase your branch you should have a place to add an entry.

Issue #551, opened 2023-02-02
Drop the custom Base64 encode and decode to favour the built-in methods


anakinj:
We currently use a custom implementation of the urlsafe encode and decode methods because switching to the `Base64.urlsafe_decode64` and `Base64.urlsafe_encode64` introduced a few issues (#483).

Think the only practical impact is that the token cannot have trailing spaces anymore.

anakinj:
Only stricter base64 decode supportedvia #658

PR #578, opened 2023-12-28
Drop use of base64 gem and warn about invalid base64 input

anakinj:
### Description

This pull request removes the dependency to the base64 gem. Motivation for this is the warnings emitting if `base64` is not in the Gemfile/gemspec of the component running it, this is something that is a thing in Ruby 3.3 forward([ref](https://github.com/ruby/ruby/blob/ruby_3_3/NEWS.md#stdlib-updates)). Another approach would be to list base64 as a dependency to in the .gemspec, but I kinda like the control of the Base64 handling. Maybe in the future we can remove the custom logic.

A bit of base64 history:

#483
#551 

This approach also allows nicer way to deprecate support for the looser base64 that is baked into RFC 2045

### Checklist


anakinj:
I changed my mind. Think it's better to add the base64 gem as a dependency.

Issue #595, opened 2024-05-21
Deprecation warning for invalid base64 should only be issued if fixed base64 is valid


jeremyevans:
Currently, passing anything that isn't strict urlsafe base64 will result in a deprecation warning.  Since this is usually passed user input, I'm not sure how one is supposed to avoid the deprecation warning, other than turning it off completely.  Ideally, the deprecation warning should only be emitted if the provided base64 is not strict urlsafe base64, but still results in a valid JWT using the loose base64 parsing.  The way it is currently implemented, you get the warning even if the base64 does not result in a JWT:

```ruby
# Does not and should not warn:
JWT.decode(JWT.encode('a', 'b'), 'b')

# Should warn and does:
JWT.decode(JWT.encode('a', 'b') + ' ', 'b')

# Warns but should not, as it raises JWT::VerificationError:
JWT.decode(JWT.encode('a', 'b') + ' 9', 'b')
```

If `JWT.decode` would raise, it should not issue a deprecation warning for invalid base64.

I realize this change requires refactoring how the deprecation warning is issued, since you would need to delay emission of the warning until after attempting parsing.

anakinj:
Very valid point, user based deprecation warnings are hard :).

I have a pretty simple idea how this could be handled, a kinda conditional deprecation message concept.

anakinj:
Fix released in 2.8.2

PR #600, opened 2024-06-16
Warn about deprecations only if no exceptions happen

anakinj:
### Description

This should fix the spammy deprecation warnings reported in #595 

### Checklist

* [ ] A description of the changes and a reference to the PR has been added to CHANGELOG.md. More details in the [CONTRIBUTING.md](https://github.com/jwt/ruby-jwt/blob/main/CONTRIBUTING.md)


anakinj:
@jeremyevans If you have some spare time could you check out this attempt to warn about the upcoming stricter base64 decoding only if the token validation is a success.

jeremyevans (commented):
I didn't test this code, but reading it, it looks like if you rescue the exception, the warning won't be emitted then, but it will be emitted later, after the next attempt that does not raise in the same thread.

I'll submit a pull request for an alternative approach and get your thoughts about it.

anakinj:
You were right, the warnings leaked over to the next call of decode. Did a little adjustment to prevent that.

jeremyevans:
> You were right, the warnings leaked over to the next call of decode. Did a little adjustment to prevent that.

My approach in #601 is simpler, does not mess with thread-local state, and uses uplevel to make it easier to find the underlying cause of the problem.  It does require base64 decoding segments twice, but as most JWTs are small, I don't expect that should be a significant performance issue.

However, either #601 or this PR will fix the problem, so I'm fine with either.

anakinj:
@jeremyevans Thanks for your interest and insight on this one. I decided to go with this more complex and hacky version. Just to avoid doing something for every token decode that goes via that method.

PR #607, opened 2024-07-30
Dynamically register algorithms to lookup as they get loaded

anakinj:
### Description

Dynamically generate the list of supported algorithms as implementations are loaded. This will allow replacing and extending the algorithm from outside of this gem.

Initial use-case would be to move the rbnacl related algorithms to a gem with an explicit dependency to the 3rd party crypto libs. More of that work can be looked at over [here](https://github.com/anakinj/jwt-eddsa)

### Checklist

### DIFF
--- a/lib/jwt/jwa.rb
+++ b/lib/jwt/jwa.rb
@@ -8,54 +8,35 @@ rescue LoadError
   raise if defined?(RbNaCl)
 end
 
-require_relative 'jwa/hmac'
-require_relative 'jwa/eddsa'
+require_relative 'jwa/signing_algorithm'
+
 require_relative 'jwa/ecdsa'
-require_relative 'jwa/rsa'
-require_relative 'jwa/ps'
+require_relative 'jwa/eddsa'
+require_relative 'jwa/hmac'
 require_relative 'jwa/none'
+require_relative 'jwa/ps'
+require_relative 'jwa/rsa'
 require_relative 'jwa/unsupported'
 require_relative 'jwa/wrapper'
 
+if JWT.rbnacl_6_or_greater?
+  require_relative 'jwa/hmac_rbnacl'
+elsif JWT.rbnacl?
+  require_relative 'jwa/hmac_rbnacl_fixed'
+end
+
 module JWT
   module JWA
-    ALGOS = [Hmac, Ecdsa, Rsa, Eddsa, Ps, None, Unsupported].tap do |l|
-      if ::JWT.rbnacl_6_or_greater?
-        require_relative 'jwa/hmac_rbnacl'
-        l << Algos::HmacRbNaCl
-      elsif ::JWT.rbnacl?
-        require_relative 'jwa/hmac_rbnacl_fixed'
-        l << Algos::HmacRbNaClFixed
-      end
-    end.freeze
-
     class << self
-      def find(algorithm)
-        indexed[algorithm&.downcase]
-      end
+      def resolve(algorithm)
+        return find(algorithm) if algorithm.is_a?(String) || algorithm.is_a?(Symbol)
 
-      def create(algorithm)
-        return algorithm if JWA.implementation?(algorithm)
-
-        Wrapper.new(*find(algorithm))
-      end
-
-      def implementation?(algorithm)
-        (algorithm.respond_to?(:valid_alg?) && algorithm.respond_to?(:verify)) ||
-          (algorithm.respond_to?(:alg) && algorithm.respond_to?(:sign))
-      end
-
-      private
-
-      def indexed
-        @indexed ||= begin
-          fallback = [nil, Unsupported]
-          ALGOS.each_with_object(Hash.new(fallback)) do |cls, hash|
-            cls.const_get(:SUPPORTED).each do |alg|
-              hash[alg.downcase] = [alg, cls]
-            end
-          end
+        unless algorithm.is_a?(SigningAlgorithm)
+          Deprecations.warning('Custom algorithms are required to include JWT::JWA::SigningAlgorithm')
+          return Wrapper.new(algorithm)
         end
+
+        algorithm
       end
     end
   end
--- a/lib/jwt/jwa/ecdsa.rb
+++ b/lib/jwt/jwa/ecdsa.rb
@@ -2,8 +2,35 @@
 
 module JWT
   module JWA
-    module Ecdsa
-      module_function
+    class Ecdsa
+      include JWT::JWA::SigningAlgorithm
+
+      def initialize(alg, digest)
+        @alg = alg
+        @digest = OpenSSL::Digest.new(digest)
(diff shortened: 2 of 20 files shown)


lib/jwt/jwa.rb
```diff
@@ -8,53 +8,37 @@
...
-          fallback = [nil, Unsupported]
-          ALGOS.each_with_object(Hash.new(fallback)) do |cls, hash|
-            cls.const_get(:SUPPORTED).each do |alg|
-              hash[alg.downcase] = [alg, cls]
-            end
-          end
+      def wrap_algorithm(algorithm)
+        return if algorithm.respond_to?(:header)
+
+        def algorithm.header(*)
```
anakinj, line 40: I wonder if it would be more clear if this would be in a separate module and extended to the given algorithm? We could maybe even use `JWT::JWA::SigningAlgorithm` directly, combined with a warning saying that algorithms needs to include this module to work in the future.

lib/jwt/jwa/ecdsa.rb
```diff
@@ -2,8 +2,36 @@
...
   module JWA
-    module Ecdsa
-      module_function
+    class Ecdsa
+      include JWT::JWA::SigningAlgorithm
+
+      def initialize(alg, digest, curve_name)
+        @alg = alg
+        @digest = OpenSSL::Digest.new(digest)
+        @curve_name = curve_name
```
anakinj, line 11: This curve_name instance variable is not used for anything, or is it?

lib/jwt/jwa/ecdsa.rb
```diff
@@ -2,8 +2,35 @@
...
+    class Ecdsa
+      include JWT::JWA::SigningAlgorithm
+
+      def initialize(alg, digest)
+        @alg = alg
+        @digest = OpenSSL::Digest.new(digest)
+      end
+
+      def sign(data:, signing_key:)
+        curve_definition = curve_by_name(signing_key.group.curve_name)
```
anakinj, line 14: Unrelated to the changes: There is not validation that the provided `signing_key` is of the expected type

lib/jwt/jwa/ecdsa.rb
```diff
@@ -2,8 +2,35 @@
...
+        key_algorithm = curve_definition[:algorithm]
+        if alg != key_algorithm
+          raise IncorrectAlgorithm, "payload algorithm is #{alg} but #{key_algorithm} signing key was provided"
```
anakinj, line 17: Unrelated to the changes: The `alg` is not something that is defined by the payload.

lib/jwt/jwa/ecdsa.rb
```diff
@@ -2,8 +2,35 @@
...
+        end
+
+        asn1_to_raw(signing_key.dsa_sign_asn1(digest.digest(data)), signing_key)
+      end
+
+      def verify(data:, signature:, verification_key:)
+        curve_definition = curve_by_name(verification_key.group.curve_name)
```
anakinj, line 24: This could be a method for validating the key and same logic used by the `sign` method.

lib/jwt/jwa/ps.rb
```diff
@@ -2,29 +2,37 @@
...
+    class Ps
       # RSASSA-PSS signing algorithms
 
-      module_function
+      include JWT::JWA::SigningAlgorithm
 
-      SUPPORTED = %w[PS256 PS384 PS512].freeze
+      def initialize(alg)
+        @alg = alg
+        @translated_algorithm = alg.sub('PS', 'sha')
```
anakinj, line 12: Lets rename this instance variable to `mgf1_hash`

lib/jwt/jwa/signing_algorithm.rb
```diff
@@ -0,0 +1,53 @@
...
+
+module JWT
+  module JWA
+    module SigningAlgorithm
+      module ClassMethods
+        def register_algorithm(algo)
+          ::JWT::JWA.register_algorithm(algo)
+        end
+      end
+
```
anakinj, line 11: Here we agreed to add stubs for verify and sign that raise if not implemented.

lib/jwt/jwa/signing_algorithm.rb
```diff
@@ -0,0 +1,53 @@
...
+      def find(algo)
+        algorithms[algo.to_s.downcase]
+      end
+
+      private
+
+      def algorithms
+        @algorithms ||= {}.tap { |h| h.default = Unsupported }
```
anakinj, line 49: We could replace the use of `default` with a fallback to the key lookup in `find`. To avoid this strange require in the middle of everything.

lib/jwt/jwa/unsupported.rb
```diff
@@ -3,16 +3,16 @@
 module JWT
   module JWA
     module Unsupported
-      module_function
+      class << self
+        include JWT::JWA::SigningAlgorithm
 
-      SUPPORTED = [].freeze
+        def sign(*)
+          raise NotImplementedError, 'Unsupported signing method'
```
anakinj, line 10: Lets not do `NotImplementedError` as it's not inherited from StandardError it can cause strangeness..

spec/jwt/jwt_spec.rb
```diff
@@ -889,6 +885,39 @@ def valid_alg?(alg)
...
+          end
+
+          def sign(*)
+            @signature
+          end
+        end
+      end
+
+      it 'emits a deprecation warning' do
+        expect { token }.to output("[DEPRECATION WARNING] Custom algorithms are required to include JWT::JWA::SigningAlgorithm\n").to_stderr
```
anakinj, line 917: Lets add a check here that the class has been extended.

lib/jwt/jwa/ecdsa.rb
anakinj: Lets improve ecdsa later

lib/jwt/jwa/ecdsa.rb
anakinj: Lets improve ecdsa later

lib/jwt/jwa/ecdsa.rb
anakinj: Lets improve ecdsa later

spec/jwt/jwt_spec.rb
anakinj: Not relevant anymore. The backwardscompatibility is achieved with the wrapper

lib/jwt/jwa/signing_algorithm.rb
anakinj: Not applicable. Resorted to a wrapper approach instead

MatteoPierro (approved):
LGTM

MatteoPierro (approved):
LGTM

PR #640, opened 2024-12-12
Handle empty string as token value

ragalie:
### Description

If the token is the empty string we try to pass `nil` to `Base64.url_decode`, which always expects a string.

This ensures we always pass a string to avoid an unexpected error.

### Checklist


anakinj (approved):

anakinj:
Nice, thanks for the fix @ragalie

Issue #562, opened 2023-05-17
Support x5t in place of kid


szh:
**Enhancement Request**

Some JWTs, including those generated by Azure DevOps, do not contain a `kid` field and instead use the `x5t` field: https://www.rfc-editor.org/rfc/rfc7515#section-4.1.7

From a cursory look through the code in this library, it seems that this should require some changes in the [key_finder.rb](https://github.com/jwt/ruby-jwt/blob/main/lib/jwt/jwk/key_finder.rb) file.

PR #669, opened 2025-04-04
Add support for x5t and x5t#S256 header

hieuk09:
### Description

Close https://github.com/jwt/ruby-jwt/issues/562

### Checklist

* [ ] A description of the changes and a reference to the PR has been added to CHANGELOG.md. More details in the [CONTRIBUTING.md](https://github.com/jwt/ruby-jwt/blob/main/CONTRIBUTING.md)


lib/jwt/decode.rb
```diff
@@ -2,6 +2,7 @@
 
 require 'json'
 require 'jwt/x5c_key_finder'
+require 'jwt/x5t_key_finder'
```
anakinj, line 5: First very high level comment is that I think i would like to move all the keyfinders to their own namespace. `JWT::KeyFinders::X5t` in this case...

lib/jwt/decode.rb
```diff
@@ -61,9 +62,12 @@ def verify_algo
     def set_key
       @key = find_key(&@keyfinder) if @keyfinder
       @key = ::JWT::JWK::KeyFinder.new(jwks: @options[:jwks], allow_nil_kid: @options[:allow_nil_kid]).key_for(token.header['kid']) if @options[:jwks]
-      return unless (x5c_options = @options[:x5c])
 
-      @key = X5cKeyFinder.new(x5c_options[:root_certificates], x5c_options[:crls]).from(token.header['x5c'])
+      if (x5c_options = @options[:x5c])
```
anakinj, line 66: If we have a namespace for the keyfinders the resolving could happen in that.

`@key = KeyFinders.resolve_key(@options, token)`

anakinj:
> @hieuk09 Did you have some reference for how this should work? Could take a look at that to educate myself a bit.

Now when looking closer at the changes I found some links...

lib/jwt/x5t_key_finder.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module JWT
+  # If the x5t header thumbprint matches one of the trusted certificates,
+  # returns the public key from that certificate.
+  # See https://tools.ietf.org/html/rfc7515#section-4.1.7 and
+  # https://tools.ietf.org/html/rfc7515#section-4.1.8
+  class X5tKeyFinder
+    def initialize(certificates)
+      raise ArgumentError, 'Certificates must be specified' unless certificates.is_a?(Array)
```
anakinj, line 10: Could also check if the array is empty and contains at least one and only the types that is expected.

lib/jwt/x5t_key_finder.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    def from(header)
+      if header['x5t']
+        x5t = header['x5t']
+        digest_class = OpenSSL::Digest::SHA1
+      elsif header['x5t#S256']
+        x5t = header['x5t#S256']
+        digest_class = OpenSSL::Digest::SHA256
+      end
```
anakinj, line 22: This feels it could be a method to extract the encoded thumbprint and digest type.

lib/jwt/x5t_key_finder.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      raise JWT::DecodeError, 'x5t or x5t#S256 header parameter is required' unless x5t
+
+      thumbprint = ::JWT::Base64.url_decode(x5t)
+      matching_cert = @certificates.find do |cert|
+        digest_class.new(cert.to_der).digest == thumbprint
```
anakinj, line 28: This is a very speculative comment:

How slow is the `#to_der` operation? I wonder if calling it for every certificate will make real-world use cases slow?

spec/jwt/x5t_key_finder_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+  let(:leaf_certificate) do
+    cert = generate_cert(leaf_dn, leaf_key.public_key, 2)
+    cert.sign(root_key, 'sha256')
+    cert
+  end
+
+  subject(:keyfinder) { described_class.new([leaf_certificate]).from(header) }
+
+  context 'when certificates argument is nil' do
+    subject(:keyfinder) { described_class.new(nil).from({}) }
```
anakinj, line 18: No need to call from in this scenario

```suggestion
    subject(:keyfinder) { described_class.new(nil) }
```

spec/jwt/x5t_key_finder_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+
+    it 'raises an argument error' do
+      expect { keyfinder }.to raise_error(ArgumentError, 'Certificates must be specified')
+    end
+  end
+
+  context 'when certificates argument is not array' do
+    subject(:keyfinder) { described_class.new('certificate').from({}) }
```
anakinj, line 26:
```suggestion
    subject(:keyfinder) { described_class.new('certificate') }
```

lib/jwt/x5t_key_finder.rb
```diff
@@ -0,0 +1,36 @@
...
+    def from(header)
+      if header['x5t']
+        x5t = header['x5t']
+        digest_class = OpenSSL::Digest::SHA1
+      elsif header['x5t#S256']
+        x5t = header['x5t#S256']
+        digest_class = OpenSSL::Digest::SHA256
+      end
+
+      raise JWT::DecodeError, 'x5t or x5t#S256 header parameter is required' unless x5t
```
anakinj, line 24: Could this message be a bit less strict as the header is not really required in general. I suggest something in the lines of "x5t or x5t#S256 header missing"

lib/jwt/x5t_key_finder.rb
anakinj: Could there be some trick to make this faster, like wrap the cert in a JWK object or something. Where the thumbprint of the JWK would be calculated once.

hieuk09:
Thank you for the detailed review. I'll take a look and fix all the comments tomorrow.

lib/jwt/decode.rb
hieuk09: That makes sense. What do you think about moving all the key finder logic to `JWT::JWK::KeyFinder`? We can abstract the logic to find key in this class. The only problem may be keeping `x5c` logic backward-compatible 🤔

lib/jwt/x5t_key_finder.rb
anakinj: I think we can leave optimizations for later iterations. Just trying to get the foundation in shape.

lib/jwt/decode.rb
hieuk09: I think I have an idea. I'll work on it tomorrow and commit. Then you can review and see whether it could work or not.

anakinj: Also if the class/module would be compatible with the new Token classes it would be cool

hieuk09:
@anakinj could you have another review?

lib/jwt/jwk/key_finder.rb
```diff
@@ -37,22 +36,36 @@ def key_for(kid)
       # Returns the key for the given token
       # @param [JWT::EncodedToken] token the token
       def call(token)
-        key_for(token.header['kid'])
+        kid = token.header['kid']
```
anakinj, line 39: If the token happens to have both a kid and for example a x5t header it will never attempt using the x5t. 

Not sure if this is a real scenario and could for sure be a later addition.

lib/jwt/jwk/rsa.rb
```diff
@@ -51,6 +51,10 @@ def verify_key
       def export(options = {})
         exported = parameters.clone
         exported.reject! { |k, _| RSA_PRIVATE_KEY_ELEMENTS.include? k } unless private? && options[:include_private] == true
+
+        exported[:x5c] = Base64.strict_encode(rsa_key.to_der) if options[:x5c]
```
anakinj, line 55: About the naming of the options, would it make sense to have the prefix `include_` on the options? To be inline with the `include_private`

anakinj, line 55: A few concerns about this.

Is the x5c header really meant to contain just a key. Isn't that supposed to have the whole certificate chain?

The header should probably be an array.

I have a feeling that `#to_der` is including the private key if such exist, probably should be specific to the public key?

anakinj (commented):
I think the JWK approach is interesting. Im a bit concerned that we are cutting some corners here, as the x5* fields are pretty certificate specific and we are currently only dealing with keypairs.

lib/jwt/jwk/key_finder.rb
hieuk09: I think this is a real scenario: here is a payload and header of token from Azure:

```json
[
  {
    "aud": "https://management.core.windows.net",
    "iss": "https://sts.windows.net/57c0dc11-c207-49dd-9bec-76a2d395ea76/",
    "iat": 1745152327,
    "nbf": 1745152327,
    "exp": 1745157876,
    "acr": "1",
    // some more data
  },
  {
    "typ": "JWT",
    "alg": "RS256",
    "x5t": "CNv0OI3RwqlHFEVnaoMAshCH2XE",
    "kid": "CNv0OI3RwqlHFEVnaoMAshCH2XE"
  }
]
```

The keys can be found in https://login.windows.net/common/discovery/keys.

lib/jwt/jwk/rsa.rb
hieuk09: That makes sense, let me update it.

hieuk09: I don't have a good idea about it because I encounter only token with `x5t` header. I think we can skip this part and revisit it later when it becomes necessary. What do you think?

lib/jwt/jwk/rsa.rb
```diff
@@ -51,6 +51,9 @@ def verify_key
       def export(options = {})
         exported = parameters.clone
         exported.reject! { |k, _| RSA_PRIVATE_KEY_ELEMENTS.include? k } unless private? && options[:include_private] == true
+
+        exported[:x5t] = Base64.url_encode(OpenSSL::Digest::SHA1.new(rsa_key.to_der).digest) if options[:include_x5t]
```
anakinj, line 55: Im still a bit uncertain if we should make the decision here what the thumbprint is, the x5t is the thumbprint for the certificate, the RSA public key is just a part of the certificate.

Think we could have a way to export a X509 certificate that could be adding these fields directly from the certificate itself.

Also with this approach the thumbprint will differ if rsa_key is a private key or just the public key.
Also what about the SHA256 version of the thumbprint?

hieuk09: That's a good point, thanks for pointing that out. Let me think a bit more about this.

anakinj: What if we would for now only focus on the key finding part? so if the JWK has a `x5*` header we are able to map them to the tokens that also have them.

hieuk09: I think the key binding should work with the code in https://github.com/jwt/ruby-jwt/pull/669/files#diff-7cb47ecf35db922bbd3b543b5df2e594c4ffc8f6bdd62c34f1b8544d6089e00fR58.

I'll remove this code (currently it's mostly used for testing anyway). What do you think?

hieuk09: Got it. Let me update the code.

lib/jwt/jwk/key_finder.rb
```diff
@@ -37,22 +36,36 @@ def key_for(kid)
...
+        x5t = token.header['x5t']
+        x5c = token.header['x5c']
+
+        if kid
+          key_for(kid, :kid)
+        elsif x5t
+          key_for(x5t, :x5t)
+        elsif x5c
```
anakinj, line 47: what about the `x5t#S256`header? Should we include it also?

lib/jwt/jwk/key_finder.rb
```diff
@@ -37,22 +36,36 @@ def key_for(kid)
...
         # First try without invalidation to facilitate application caching
-        @jwks ||= JWT::JWK::Set.new(@jwks_loader.call(kid: kid))
+        @jwks ||= JWT::JWK::Set.new(@jwks_loader.call(key_field => kid))
         jwk = @jwks.find { |key| key_matcher.call(key) }
 
         return jwk if jwk
 
         # Second try, invalidate for backwards compatibility
-        @jwks = JWT::JWK::Set.new(@jwks_loader.call(invalidate: true, kid_not_found: true, kid: kid))
+        @jwks = JWT::JWK::Set.new(@jwks_loader.call(invalidate: true, kid_not_found: true, key_field => kid))
```
anakinj, line 68: This could cause some problems as the keyloader could always be expecting the kid. Also do we will start invalidating 3 times if nothing is found.

anakinj: Also the line above where we do `call` with the `key_field` could cause backwards compatibility issues.

hieuk09: > This could cause some problems as the keyloader could always be expecting the kid 

I think it's fine because we only call the JWKS loader with other key fields if token header does not contain `kid`. I think most of the JWKS endpoints support `kid` by default.

> Also do we will start invalidating 3 times if nothing is found.

No, we only invalidating once. Basically, if `kid` field exists in token header, we only match using `kid`. Otherwise, if `x5t` exists and there is no `kid`, we'll match using `x5t`.

> Also the line above where we do call with the key_field could cause backwards compatibility issues.

Could you clarify on what could be the compatibility issues? I change `key_for` signature but I added `key_field` as optional parameter so it should be backward-compatible. I may miss other scenarios, so please let me know.

lib/jwt/jwk/key_finder.rb
hieuk09: I don't have a good idea how `x5t#S256` will be returned in JWK params (the payload should contain a field `x5t#S256` but I wonder how will that be parsed) so I don't add that for now. I could add it quickly if you think it's better to add that.

anakinj: Ok, if it's unclear lets leave it out.

lib/jwt/jwk/key_finder.rb
anakinj: You're right! It happens only once. Now the gem is dictating the lookup order, so if the key contains both kid and x5 headers we'll never try the x5 lookups. 

I feel this key lookup is already so complicated, cant wrap my head around this.

lib/jwt/jwk/key_finder.rb
```diff
@@ -37,22 +36,36 @@ def key_for(kid)
...
       private
 
-      def resolve_key(kid)
-        key_matcher = ->(key) { (kid.nil? && @allow_nil_kid) || key[:kid] == kid }
+      def resolve_key(kid, key_field)
+        key_matcher = ->(key) { (kid.nil? && @allow_nil_kid) || key[key_field] == kid }
 
         # First try without invalidation to facilitate application caching
-        @jwks ||= JWT::JWK::Set.new(@jwks_loader.call(kid: kid))
+        @jwks ||= JWT::JWK::Set.new(@jwks_loader.call(key_field => kid))
```
anakinj, line 62: key_field should probably be set to `:kid` at before this is called if `@allow_nil_kid` is true

anakinj: I wonder if it's too late to drop the @allow_nil_kid option for the 3.0 version... 🤔

anakinj: Would simplify this complicated piece of machinery....

lib/jwt/jwk/key_finder.rb
hieuk09: That's right, the gem is dictating the lookup order. Currently I couldn't think of a better way to do this. Do you have any suggestions?

anakinj: For use via the `JWT.decode` method the predefined order is fine i think.

What about having the order as a option when initializing the keyfinder with a sensible default.

Then the new token object approach could be easily customized.
```
key_finder = JWT::JWK::KeyFinder.new(jwks: JWT::JWK::Set.new(jwk), key_fields: [:kid, :x5t])
encoded_token = JWT::EncodedToken.new(token.jwt)
encoded_token.verify!(signature: { algorithm: 'RS256', key_finder: key_finder})
```

hieuk09: Thanks, that's a good idea. Let me work on that.

hieuk09: This is done in 08d3bdddf65fd1606cb711faa319108956afee4a

hieuk09:
@anakinj I think there are no pending tasks for this PR, the key field should be totally optional now.

PR #697, opened 2025-06-27
Avoid using the same digest across calls

headius:
### Description

JWT appears to reuse these JWA instances across threads, which can lead to them stepping on each other via the shared OpenSSL::Digest instance. This causes decoding to fail verification, likely because the digest contains an amalgam of data from the different threads.

This patch creates a new OpenSSL::Digest for each use, avoiding the threading issue.

Note that the HMAC JWA already calls OpenSSL::HMAC.digest, avoiding the shared state, and the others do not use digest.

The original code does not fail on CRuby most likely because only one thread at a time can be calculating a digest against a given OpenSSL::Digest instance, due to the VM lock.

Fixes https://github.com/jwt/ruby-jwt/issues/696

Addresses the issue reported in https://github.com/jruby/jruby/issues/8504 by @mohamedhafez

### Checklist

- [ ] There are tests for the fix or feature added/changed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 **Fixes and enhancements:**
 
-- Your contribution here
+- Avoid using the same digest across calls in JWT::JWA::Ecdsa and JWT::JWA::Rsa [#697](https://github.com/jwt/ruby-jwt/pull/697)
 
 ## [v3.1.1](https://github.com/jwt/ruby-jwt/tree/v3.1.1) (2025-06-24)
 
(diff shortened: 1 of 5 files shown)


headius:
Some variation of the script provided in #696 could probably be adapted as a test for the ECDSA part of the fix. I am not familiar enough with the libraries involved to know how to modify that script to test the RSA part of the fix.

headius:
I've adapted the #696 example for an ECDSA test.

anakinj:
Re-enabled the Truffleruby CI step in #698. I can confirm that the provided spec is indeed failing for that.

headius:
Well, I guess I answered my own question... there are just a few failures running on JRuby, and something that appears to hang. I will look into those separately from this PR.

I've made the Rubocop changes and this should now be ready to go.

anakinj (approved):

PR #712, opened 2026-01-29
Implements Nested JWT functionality as defined in RFC 7519 Section 5.2, 7.1, 7.2, and Appendix A.2.

ydah:
### Description

Implements Nested JWT functionality as defined in RFC 7519 Section 5.2, 7.1, 7.2, and Appendix A.2.

A Nested JWT is a JWT used as the payload of another JWT, allowing multiple layers of signing with different keys/algorithms.

see: https://datatracker.ietf.org/doc/html/rfc7519#section-5.2
see: https://datatracker.ietf.org/doc/html/rfc7519#section-7.1
see: https://datatracker.ietf.org/doc/html/rfc7519#section-7.2

### Checklist

### DIFF
--- a/lib/jwt/encoded_token.rb
+++ b/lib/jwt/encoded_token.rb
@@ -13,7 +13,7 @@ module JWT
   #   encoded_token = JWT::EncodedToken.new(token.jwt)
   #   encoded_token.verify_signature!(algorithm: 'HS256', key: 'secret')
   #   encoded_token.payload # => {'pay' => 'load'}
-  class EncodedToken
+  class EncodedToken # rubocop:disable Metrics/ClassLength
     DEFAULT_CLAIMS = [:exp].freeze
 
     private_constant(:DEFAULT_CLAIMS)
@@ -205,7 +205,7 @@ module JWT
     def inner_token
       return nil unless nested?
 
-      EncodedToken.new(unverified_payload)
+      EncodedToken.new(decode_nested_payload)
     end
 
     # Unwraps all nesting levels and returns an array of tokens.
@@ -218,11 +218,13 @@ module JWT
     #   all_tokens = token.unwrap_all
     #   all_tokens.first # => outermost token
     #   all_tokens.last  # => innermost token
-    def unwrap_all
+    def unwrap_all(max_depth:)
       tokens = [self]
       current = self
 
       while current.nested?
+        raise JWT::DecodeError, "Nested JWT exceeds maximum depth of #{max_depth}" if tokens.length >= max_depth
+
         current = current.inner_token
         tokens << current
       end
@@ -247,6 +249,14 @@ module JWT
       end
 
       parse_and_decode(encoded_payload)
+    end
+
+    def decode_nested_payload
+      raise JWT::DecodeError, 'Encoded payload is empty' if encoded_payload == ''
+
+      return encoded_payload if unencoded_payload?
+
+      ::JWT::Base64.url_decode(encoded_payload || '')
     end
 
     def unencoded_payload?
--- a/lib/jwt/encoded_token.rb
+++ b/lib/jwt/encoded_token.rb
@@ -178,6 +178,58 @@ module JWT
 
     alias to_s jwt
 
+    # Checks if this token is a Nested JWT.
+    # A token is considered nested if it has a `cty` header with value "JWT" (case-insensitive).
+    #
+    # @return [Boolean] true if this is a Nested JWT, false otherwise
+    #
+    # @example
+    #   token = JWT::EncodedToken.new(nested_jwt_string)
+    #   token.nested? # => true
+    #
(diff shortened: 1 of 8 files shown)


anakinj:
Thanks for the addition @ydah, looks interesting. I need to take a look at this with a fresh mind. But its on my todo list.

lib/jwt/encoded_token.rb
```diff
@@ -11,7 +11,7 @@ module JWT
   #   encoded_token = JWT::EncodedToken.new(token.jwt)
   #   encoded_token.verify_signature!(algorithm: 'HS256', key: 'secret')
   #   encoded_token.payload # => {'pay' => 'load'}
-  class EncodedToken
+  class EncodedToken # rubocop:disable Metrics/ClassLength
```
anakinj, line 16: Could we do the refactoring sketched out in #713 in a separate PR to deal with the class length?

lib/jwt/encoded_token.rb
```diff
@@ -192,6 +192,58 @@ def valid_claims?(*options)
...
+    # @example
+    #   token = JWT::EncodedToken.new(deeply_nested_jwt)
+    #   all_tokens = token.unwrap_all
+    #   all_tokens.first # => outermost token
+    #   all_tokens.last  # => innermost token
+    def unwrap_all
+      tokens = [self]
+      current = self
+
+      while current.nested?
```
anakinj, line 225: Should we limit the depth of the nesting to for example 10? There could be some security issue with having no limit.

lib/jwt/token.rb
```diff
@@ -127,5 +127,31 @@ def valid_claims?(*options)
     #
     # @return [String] the JWT token as a string.
     alias to_s jwt
+
+    class << self
```
anakinj: Would be easier to follow if there would be only one way to create these for now. Fancying smaller iterations over different variations how to do things.

lib/jwt/nested_token.rb
```diff
@@ -0,0 +1,111 @@
...
+  #     ]
+  #   )
+  #   inner_payload = tokens.last.payload
+  #
+  # @see https://datatracker.ietf.org/doc/html/rfc7519#section-5.2 RFC 7519 Section 5.2
+  class NestedToken
+    # The content type header value for nested JWTs as per RFC 7519
+    CTY_JWT = 'JWT'
+
+    class << self
```
anakinj: Could we represent such a NestedToken as an instance instead of having static methods doing everything (signing, claim/signature, verification)?

Now for example there is no way to control what claims are being verified.

lib/jwt/nested_token.rb
```diff
@@ -0,0 +1,111 @@
...
+      # @example With RSA and custom headers
+      #   nested = JWT::NestedToken.sign(
+      #     inner_jwt,
+      #     algorithm: 'RS256',
+      #     key: rsa_private_key,
+      #     header: { kid: 'my-key-id' }
+      #   )
+      def sign(inner_jwt, algorithm:, key:, header: {})
+        outer_header = header.merge('cty' => CTY_JWT)
+        token = Token.new(payload: inner_jwt, header: outer_header)
```
anakinj: Think we will be creating invalid tokens this way, as `Token` is always doing the JSON encoding on the payload. To my understanding the nested tokens should only be base64 encoded.

lib/jwt/encoded_token.rb
ydah: Indeed it is! We've limited it to 10.

lib/jwt/token.rb
ydah: I've organized and refined it. What do you think of this?

lib/jwt/nested_token.rb
ydah: I've represented this series of processes as an instance. What do you think of this?

lib/jwt/nested_token.rb
ydah: Indeed it is. I fixed it.

lib/jwt/encoded_token.rb
```diff
@@ -13,7 +13,7 @@ module JWT
   #   encoded_token = JWT::EncodedToken.new(token.jwt)
   #   encoded_token.verify_signature!(algorithm: 'HS256', key: 'secret')
   #   encoded_token.payload # => {'pay' => 'load'}
-  class EncodedToken
+  class EncodedToken # rubocop:disable Metrics/ClassLength
```
anakinj, line 16: Would be super nice if we could avoid polluting the EncodedToken with the nested token logic. Could we have a pair: `NestedToken` and `EncodedNestedToken`

anakinj: I vibed together a rough sketch of what I'm visualizing: https://github.com/jwt/ruby-jwt/pull/721

PR #713, opened 2026-01-29
Add duplicate claim name detection per RFC 7519 Section 4

ydah:
### Description

Implements duplicate claim name detection as specified in RFC 7519 Section 4, which states:

> The Claim Names within a JWT Claims Set MUST be unique; JWT parsers MUST either reject JWTs with duplicate Claim Names or use a JSON parser that returns only the lexically last duplicate member name.

see: https://datatracker.ietf.org/doc/html/rfc7519#section-4

This feature allows users to reject JWTs that contain duplicate keys in the header or payload, which is recommended for security-sensitive applications to prevent claim confusion attacks.

### Checklist


lib/jwt/json.rb
```diff
@@ -1,18 +1,129 @@
...
+      #   # => raises JWT::DuplicateKeyError
+      def parse(data, allow_duplicate_keys: true)
+        DuplicateKeyChecker.check!(data) unless allow_duplicate_keys
         ::JSON.parse(data)
       end
     end
+
+    # @api private
+    # Checks for duplicate keys in a JSON string using a StringScanner-based tokenizer
+    # rubocop:disable Style/RedundantRegexpArgument
```
ydah: see: https://github.com/jwt/ruby-jwt/actions/runs/21479506016/job/61871895439

lib/jwt/encoded_token.rb
```diff
@@ -39,11 +39,14 @@ def payload
     # Initializes a new EncodedToken instance.
     #
     # @param jwt [String] the encoded JWT token.
+    # @param allow_duplicate_keys [Boolean] whether to allow duplicate keys in header/payload (default: true).
     # @raise [ArgumentError] if the provided JWT is not a String.
-    def initialize(jwt)
+    # @raise [JWT::DuplicateKeyError] if allow_duplicate_keys is false and duplicate keys are found.
+    def initialize(jwt, allow_duplicate_keys: true)
```
anakinj: Would like to avoid bloating the initializer with feature parameters, the original `JWT.encode` and `JWT.decode` methods became parameter monsters over time.

A few random options, not sure if Im a fan of either of my own suggestions :)

```
token = EncodedToken.new(jwt)
token.raise_on_duplicate_keys!
```

```
token = EncodedToken.new(jwt)
token.decode_payload!(raise_on_duplicate_keys:true)
```

Or allowing changing the default JSON decoder with a stricter one, `token.raise_on_duplicate_keys!` could internally do something like this.

```
token = EncodedToken.new(jwt)
token.parser = JWT::JSON.new(raise_on_duplicate_keys:true)
```

lib/jwt/json.rb
```diff
@@ -1,18 +1,129 @@
...
+      # @raise [JWT::DuplicateKeyError] if allow_duplicate_keys is false and duplicate keys are found
+      #
+      # @example Default behavior (allows duplicates, uses last value)
+      #   JWT::JSON.parse('{"a":1,"a":2}') # => {"a" => 2}
+      #
+      # @example Strict mode (rejects duplicates)
+      #   JWT::JSON.parse('{"a":1,"a":2}', allow_duplicate_keys: false)
+      #   # => raises JWT::DuplicateKeyError
+      def parse(data, allow_duplicate_keys: true)
+        DuplicateKeyChecker.check!(data) unless allow_duplicate_keys
```
anakinj: Should we try to use the built-in-soon-to-be-default `::JSON.parse(data, allow_duplicate_key: false)` to deal with this issue. Feels like this gem is not supposed to be dealing with parsing concerns.

Also the new default will kick in in json 3.0. Think if the user has not taken a stance the default behavior of the json gem should be honored.

So adding as little logic as possible related to this feature will make future maintenance easier.

lib/jwt/decode.rb
```diff
@@ -119,5 +119,11 @@ def none_algorithm?
     def alg_in_header
       token.header['alg']
     end
+
+    def allow_duplicate_keys?(options)
```
anakinj: I have the urge not to add features to the old API anymore. Could we try that and document how to use this feature using the new `EncodedToken` class.

lib/jwt/configuration/decode_configuration.rb
```diff
@@ -34,7 +36,8 @@ class DecodeConfiguration
                     :verify_sub,
                     :leeway,
                     :algorithms,
-                    :required_claims
+                    :required_claims,
+                    :allow_duplicate_keys
```
anakinj: Can we live without a configuration option for this?

anakinj:
Overall I support this feature and think it should be the default behavior.

My biggest concern is adding a lot of logic to this now it will stay and live in the gem in the future. 

Would almost just like to require `json >= 2.13.0` and always parse the json with `allow_duplicate_key: false`. Then ship a jwt gem with this potentially breaking change. If needed we can then add a feature to loosen this behavior by for example allowing swapping the JSON parser.

lib/jwt/configuration/decode_configuration.rb
ydah: OK, delete it.

lib/jwt/decode.rb
ydah: I've added instructions for use. How does that look?

lib/jwt/encoded_token.rb
ydah: I've organized it. What do you think?

.github/workflows/test.yml
```diff
@@ -33,8 +33,6 @@ jobs:
         os:
           - ubuntu-latest
         ruby:
-          - "2.5"
```
ydah, line 36: MEMO: We added json gem >= 2.13.0 to the dependencies, but this version requires Ruby 2.7+. Since ruby-jwt has required_ruby_version = ‘>= 2.5’, dependency resolution fails during bundle install in a Ruby 2.5 environment.

ydah:
I updated this PR. WDYT?

lib/jwt/json.rb
```diff
@@ -3,15 +3,36 @@
 require 'json'
 
 module JWT
+  # JSON parsing utilities with duplicate key detection support
   # @api private
   class JSON
     class << self
+      # Generates a JSON string from the given data
```
anakinj: As these apis are marked private and pretty self-explanatory I would not bother having detailed comments on them.

lib/jwt/json.rb
```diff
@@ -3,15 +3,36 @@
...
+      # @raise [JWT::DuplicateKeyError] if allow_duplicate_keys is false and duplicate keys are found
+      #
+      # @example Default behavior (allows duplicates, uses last value)
+      #   JWT::JSON.parse('{"a":1,"a":2}') # => {"a" => 2}
+      #
+      # @example Strict mode (rejects duplicates)
+      #   JWT::JSON.parse('{"a":1,"a":2}', allow_duplicate_keys: false)
+      #   # => raises JWT::DuplicateKeyError
+      def parse(data, allow_duplicate_keys: true)
+        ::JSON.parse(data, allow_duplicate_key: allow_duplicate_keys)
```
anakinj: This is problematic in a way that when JSON 3.0 ships with the switched default we still are allowing duplicate keys by default. Would rather at that point enforce the new default for everyone.

Could we pass the parameter to the JSON lib only if the parameter is `false`,

lib/jwt/encoded_token.rb
```diff
@@ -44,12 +44,29 @@ def initialize(jwt)
...
+    #
+    # @example
+    #   token = JWT::EncodedToken.new(jwt_string)
+    #   token.raise_on_duplicate_keys!
+    #   token.header # May raise JWT::DuplicateKeyError
+    #
+    # @return [self]
+    # @raise [JWT::DuplicateKeyError] if duplicate keys are found during subsequent parsing.
+    def raise_on_duplicate_keys!
+      @allow_duplicate_keys = false
```
anakinj, line 67: Could we here detect if the JSON gem is supporting the feature and raise if we are not compatible? Would allow us to support older Ruby versions still.

lib/jwt/encoded_token.rb
```diff
@@ -11,7 +11,7 @@ module JWT
   #   encoded_token = JWT::EncodedToken.new(token.jwt)
   #   encoded_token.verify_signature!(algorithm: 'HS256', key: 'secret')
   #   encoded_token.payload # => {'pay' => 'load'}
-  class EncodedToken
+  class EncodedToken # rubocop:disable Metrics/ClassLength
```
anakinj: Wonder if we could do something to not need to disable cops 🤔

.github/workflows/test.yml
anakinj: Would like to support as many versions as possible, what about dynamically figure out if its possible to enable this feature or not.

ydah: Indeed, I've restored support for 2.5 and 2.6.

lib/jwt/json.rb
ydah: Indeed, I delete it.

lib/jwt/json.rb
ydah: I fixed it. Please point out if I understand something wrong.

lib/jwt/encoded_token.rb
ydah: It sounds good! I modified it that way. WDYT?

lib/jwt/encoded_token.rb
ydah: I've extracted some classes. WDYT?

CHANGELOG.md
```diff
@@ -4,9 +4,13 @@
 
 [Full Changelog](https://github.com/jwt/ruby-jwt/compare/v3.1.2...v3.1.3)
 
+**Breaking changes:**
+- Drop support for Ruby 2.6 and older [#713](https://github.com/jwt/ruby-jwt/pull/713) ([@ydah](https://github.com/ydah))
+- Bump minimum json gem version to 2.6 [#713](https://github.com/jwt/ruby-jwt/pull/713) ([@ydah](https://github.com/ydah))
+
```
anakinj, line 10:
```suggestion

```

No breaking changes anymore, right?

lib/jwt/encoded_token.rb
anakinj: The refactorings are pretty nice but could we do those in a separate PR?

anakinj:
Sorry for the back and forth but im a bit starting to regret my suggestions to only allowing enabling it per instance. I wonder if it would still be easier for most users to enable this globally for the gem. Also setting the flag to true/false would allow disabling of the strictness when JSON 3.0 is out with the new default.

What do you think:
```
JWT.configuration.decode.allow_duplicate_json_keys 

nil/not set - default behavior of the JSON gem
false - always deny, if set require the JSON gem that supports it
true - always allow, no gem restrictions
```

Also as mentioned earlier maybe do the refactorings in separate to keep the scope of this change on the new feature?

PR #715, opened 2026-02-02
Fix type error when header is not a JSON object

304:
### Description

Prevent TypeError when decoding tokens with non-hash headers (e.g., arrays).

Fixes a TypeError where malformed tokens with array headers caused "no implicit conversion of String into Integer" error instead of raising the expected `JWT::DecodeError`.

### Checklist

### DIFF
--- a/spec/jwt/jwt_spec.rb
+++ b/spec/jwt/jwt_spec.rb
@@ -8,6 +8,7 @@ RSpec.describe JWT do
       :empty_token => 'e30K.e30K.e30K',
       :empty_token_2_segment => 'e30K.e30K.',
       :invalid_header_token => 'W10.e30K.e30K',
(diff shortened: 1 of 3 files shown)


spec/jwt/jwt_spec.rb
```diff
@@ -472,6 +473,14 @@
           end
         end
 
+        context 'invalid 2-segment header format' do
+          it 'should raise JWT::DecodeError' do
+            expect do
+              JWT.decode data[:invalid_2_segment_header_token]
+            end.to raise_error JWT::DecodeError
```
anakinj: Could we check that the error message or that a part of it is the expected message? The `JWT::DecodeError` could basically be any of the errors raised.

304: Thanks for the review, @anakinj! I’ve expanded the test to also check the message.

anakinj (approved):

PR #716, opened 2026-02-14
Add enforce_hmac_key_length configuration option

304:
### Description

Add `enforce_hmac_key_length` configuration option to enforce minimum key lengths for HMAC algorithms. This is a security hardening measure that helps prevent brute-force attacks on weak keys.

Recommended minimum key lengths per RFC 7518 Section 3.2:
- HS256: 256 bits (32 bytes)
- HS384: 384 bits (48 bytes)
- HS512: 512 bits (64 bytes)

See: https://www.rfc-editor.org/rfc/rfc7518.html#section-3.2

Usage:
```ruby
  JWT.configuration.decode.enforce_hmac_key_length = true
```

For backward compatibility, it's disabled by default (false). But it would be nice to enforce it in a future major version like it's done for RSA key validation (minimum 2048 bits). For now, users must explicitly opt-in.

### Checklist

### DIFF
--- a/lib/jwt/configuration/decode_configuration.rb
+++ b/lib/jwt/configuration/decode_configuration.rb
@@ -24,6 +24,8 @@ module JWT
       #   @return [Array<String>] the list of acceptable algorithms.
       # @!attribute [rw] required_claims
       #   @return [Array<String>] the list of required claims.
+      # @!attribute [rw] enforce_hmac_key_length
+      #   @return [Boolean] whether to enforce minimum HMAC key lengths. false disables validation (default).
 
       attr_accessor :verify_expiration,
                     :verify_not_before,
@@ -34,7 +36,8 @@ module JWT
                     :verify_sub,
                     :leeway,
                     :algorithms,
-                    :required_claims
+                    :required_claims,
+                    :enforce_hmac_key_length
 
       # Initializes a new DecodeConfiguration instance with default settings.
       def initialize
@@ -48,6 +51,7 @@ module JWT
         @leeway = 0
         @algorithms = ['HS256']
         @required_claims = []
+        @enforce_hmac_key_length = false
       end
 
       # @api private
@@ -62,7 +66,8 @@ module JWT
           verify_sub: verify_sub,
           leeway: leeway,
           algorithms: algorithms,
-          required_claims: required_claims
+          required_claims: required_claims,
+          enforce_hmac_key_length: enforce_hmac_key_length
         }
       end
     end
--- a/lib/jwt/jwa/hmac.rb
+++ b/lib/jwt/jwa/hmac.rb
@@ -6,6 +6,15 @@ module JWT
     class Hmac
       include JWT::JWA::SigningAlgorithm
 
+      # Minimum key lengths for HMAC algorithms based on RFC 7518 Section 3.2.
+      # Keys must be at least the size of the hash output to ensure sufficient
+      # entropy for the algorithm's security level.
+      MIN_KEY_LENGTHS = {
+        'HS256' => 32,
+        'HS384' => 48,
+        'HS512' => 64
+      }.freeze
+
       def initialize(alg, digest)
         @alg = alg
         @digest = digest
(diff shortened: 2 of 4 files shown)


lib/jwt/configuration/decode_configuration.rb
```diff
@@ -62,7 +66,8 @@ def to_h
           verify_sub: verify_sub,
           leeway: leeway,
           algorithms: algorithms,
-          required_claims: required_claims
+          required_claims: required_claims,
+          hmac_min_key_length: hmac_min_key_length
```
anakinj: Could we instead of letting users set the min length just let the gem validate against the recommended minimums, so just enable or disable the feature. Then in future versions we could enable it by default. Something like `enforce_hmac_key_length`

I think this would be a bit less complex and easier for users to toggle on.

lib/jwt/jwa/hmac.rb
```diff
@@ -6,6 +6,12 @@ module JWA
     class Hmac
       include JWT::JWA::SigningAlgorithm
 
+      MIN_KEY_LENGTHS = {
```
anakinj, line 12: Future us could wonder where are these values coming from, a comment pointing to the RFC would be useful at this point.

lib/jwt/configuration/decode_configuration.rb
anakinj: Or is there a use-case where you would like to enforce a higher minimum than the default...

304: Good point! I initially considered making it configurable, but default minimums should be sufficient.

I’ve pushed new changes to the branch using the `enforce_hmac_key_length` boolean flag instead.

Thanks for the review, @anakinj! Could you please take another look?

anakinj (approved):
Great stuff, thanks for the addition. Next major release should swap the default to `true`

lib/jwt/jwa/hmac.rb
```diff
@@ -34,6 +50,15 @@ def verify(data:, signature:, verification_key:)
 
       attr_reader :digest
 
+      def validate_key_length!(key)
+        return unless JWT.configuration.decode.enforce_hmac_key_length
+
+        min_length = MIN_KEY_LENGTHS[alg]
+        return if key.bytesize >= min_length
```
nevans, line 57: This should guard against `nil` to future proof against unknown algorithms, right?

Should it be permissive for unknown algorithms?

```suggestion
        min_length = MIN_KEY_LENGTHS[alg]
        return if minkey_length &.<= key.bytesize
```

IMO, that's not appropriate.  It just leaves open a vector for potential security vulnerabilities.

Should it enforce that new algorithms are also added to `MIN_KEY_LENGTHS`?

```suggestion
        min_length = MIN_KEY_LENGTHS.fetch(alg)
        return if key.bytesize >= min_length
```

Alternatively, you can just ask the digest what its length should be:
```suggestion
        min_length = MIN_KEY_LENGTHS.fetch(alg) { digest.new.digest_length }
        return if key.bytesize >= min_length
```

You could also avoid the extra allocation by instantiating `digest.new` prior to calling `validate_key_length!` and pass in `min_length: digest_obj.digest_length` as an argument.  Then you could also simplify it so the `MIN_KEY_LENGTHS` hash is completely unnecessary:

```suggestion
        return if key.bytesize >= min_length
```

I'd personally vote for that last option.

lib/jwt/jwa/hmac.rb
```diff
@@ -34,6 +50,15 @@ def verify(data:, signature:, verification_key:)
...
+
+        raise_verify_error!("HMAC key must be at least #{min_length} bytes for #{alg} algorithm")
```
nevans, line 59: IMO, although it's reasonable to preserve backwards compatibility, this should at least warn by default.  This may be insignificant in development or CI, but it could be very significant in production.  Also, I always appreciate when breaking changes are preceded by a warning period.

Personally, I'd use `enforce_hmac_key_length = :warn` (vs other truthy values) for warnings.  But you could also use `nil` to indicate unset/default (to warn instead of raise) vs `false` for explicitly set (to opt-out of warnings).


```suggestion
        msg = "HMAC key must be at least #{min_length} bytes for #{alg} algorithm"
        if JWT.configuration.decode.enforce_hmac_key_length == :warn
          warn msg
        else
          raise_verify_error!(msg)
        end
```

nevans (commented):
This is great, thanks!  I have a few "drive by" comments.  If you think they're reasonable, I can convert them into a PR.

anakinj:
@nevans Your suggestions sounds reasonable. I'd be happy to see a PR with the improvements.

lib/jwt/jwa/hmac.rb
anakinj: If you have time to look into these I would suggest we swap the default to warn already in the next release.

PR #728, opened 2026-05-27
Ignore unkown algorithm jwks per RFC7517

ewindisch:
According to RFC7517, unknown algorithms in JWKS should be ignored.

Adherence to this property allow for algorithm upgrade and rotation, and notably is leveraged by existing draft RFC for hybrid PQC. Wherein, a classical algorithm such as RS256 may be used, in addition to the new AKP-type algorithm. All known algorithms must be verified, but all unknown algorithms may be skipped, per RFC7517.

I have tested popular libraries and ruby-jwt is an outlier in its behavior:

| Library | Language | Behavior with unknown kty: "AKP" in JWKS | Result |
|---|---|---|---|
| jose v5.10 | Node.js | Skips unknown key, matches kid to RSA key | PASS |
| PyJWT v2.13 | Python | PyJWKClient filters unknown keys, proceeds | PASS |
 | nimbus-jose-jwt | Java (Spring) | Silently ignores unsupported keys | PASS |
 | go-jose | Go | Returns ErrUnsupportedKeyType per-key, skips it | PASS |
 | ruby-jwt | Ruby | Throws JWT::JWKError: Key type AKP not supported | FAIL |

This change causes ruby-jwt to ignore unknown key types, and enables it to verify keys from signers who have adopted PQC type algorithms such as AKP as part of a hybrid scheme.

### Description

This Pull Request changes/fixes this thing

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## [v3.2.1](https://github.com/jwt/ruby-jwt/tree/v3.2.1) (NEXT)
+
+[Full Changelog](https://github.com/jwt/ruby-jwt/compare/v3.2.0...v3.2.1)
+
+**Features:**
+
+- Your contribution here
+
+**Fixes and enhancements:**
+
+- Fix rejection of unknown algorithms from JWKs for RFC compliance and pquip [#728](https://github.com/jwt/ruby-jwt/pull/728)
+
 ## [v3.2.0](https://github.com/jwt/ruby-jwt/tree/v3.2.0) (2026-05-13)
 
 [Full Changelog](https://github.com/jwt/ruby-jwt/compare/v3.1.2...v3.2.0)
--- a/lib/jwt/error.rb
+++ b/lib/jwt/error.rb
@@ -51,4 +51,7 @@ module JWT
 
   # The JWKError class is raised when there is an error with the JSON Web Key (JWK).
   class JWKError < DecodeError; end
+
+  # Raised when a JWK uses a key type (kty) that this library does not support.
+  class UnsupportedKeyType < JWKError; end
 end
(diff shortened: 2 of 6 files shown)


anakinj:
Nice catch and fix. Took a while to find the the reference in the RFC. I guess you are pointing to this statement:
> Additional members can be present in the JWK Set; if not understood
> by implementations encountering them, they MUST be ignored.

There seems to be some Ruby 2.5 incompatibilities, as it should be trivial to support outdated rubies I still think its valuable to keep the support.

ewindisch:
Great. I'll fix the Ruby 2.5 support and update the PR.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Fix compatibility with the openssl 4.0 gem [#706](https://github.com/jwt/ruby-jwt/pull/706)
 - Test against Ruby 4.0 on CI [#707](https://github.com/jwt/ruby-jwt/pull/707)
 - Fix type error when header is not a JSON object [#715](https://github.com/jwt/ruby-jwt/pull/715) - ([@304](https://github.com/304))
+- Fix rejection of unknown algorithms from JWKs for RFC compliance and pquip [#728](https://github.com/jwt/ruby-jwt/pull/728)
```
anakinj: Oh, I've apparently forgot to start the next iteration after the security fix last week. The changelog entry would need to go into that when its there.

anakinj: This is now done and it landed on `main`. You could move your entry to the next release in the changelog

ewindisch: Done. I was confused what to do in regard to editing the CHANGELOG, I appreciate the clarity.

anakinj (approved):

anakinj:
Thank you for noticing and going the extra mile to fix the behavior. 🙌

PR #730, opened 2026-05-27
Add step to install dependencies in release workflow

connorshea:
### Description

Sorry, I made a mistake in the previous PR because I had assumed that `bundler-cache: false` would still install dependencies. 

Apparently, it does not. So we have to install them explicitly (while continuing to not do caching, to protect against cache poisoning).

### Checklist


anakinj (approved):

anakinj:
Thanks. I'll know to keep an eye on this on the next release.

Issue #743, opened 2026-07-24
JWT::JWK::Set raises on an unknown `kty`, failing the whole set (RFC 7517 §5)


RalphBragg:
### Summary

Building a `JWT::JWK::Set` from a JWKS that contains a key with an unrecognised `kty` raises `JWT::JWKError: Key type AKP not supported`, so the whole set fails. Per RFC 7517 section 5 an unrecognised key should be ignored, not fatal, so one post-quantum key in a published JWKS takes down verification of the classical keys with it.

### Reproduction

```ruby
require "jwt"
JWT::JWK::Set.new(mixed_jwks)   # RSA + one {"kty" => "AKP", ...} (ML-DSA, RFC 9964)
# JWT::JWKError: Key type AKP not supported
```

Measured against ruby-jwt 3.2.0. The set builds fine once the AKP key is removed. Fixture and probe available on request.

### Why this matters

RFC 7517 section 5: *"Implementations SHOULD ignore JWKs within a JWK Set that use `kty` values that are not understood by them."* ML-DSA (FIPS 204 / RFC 9964, `kty: "AKP"`) is starting to appear in published key sets as post-quantum migration begins. When an issuer adds one to a JWKS that ruby-jwt consumers read, those consumers stop verifying every signature from that issuer, including classical RS256/ES256, on a spec-compliant key publication they do not control. In federated deployments that is an externally triggered outage.

### Suggested fix

When constructing a `JWT::JWK::Set`, skip a member whose `kty` is unsupported (optionally surface it) rather than raising, so the recognised keys still load and remain usable for `kid` lookup. This implements the RFC section 5 SHOULD and leaves recognised keys unchanged.

### Offer

Happy to share the reproduction harness (mixed classical + AKP JWKS plus a control) and to open a PR with the per-key skip and a test.

RalphBragg:
Update after digging in: on `main` the Hash constructor already ignores unknown `kty` per RFC 7517 §5 (fixed in #728), so `JWT::JWK::Set.new({keys: [...]})` is tolerant there. The remaining gap is the **array** constructor, `JWT::JWK::Set.new([...])`, which still fails the whole set on one unknown key. Opened #744 to make the array path consistent with the hash path (with a mirrored spec). Thanks!

PR #744, opened 2026-07-24
Ignore unsupported kty when a Set is built from an array (RFC 7517 §5)

RalphBragg:
### What

Makes `JWT::JWK::Set.new(array_of_keys)` skip a key whose `kty` it does not recognise, instead of raising `JWT::UnsupportedKeyType` and failing the whole set.

### Why

The Hash constructor already does this (added in #728, *"Ignore unknown algorithm jwks per RFC 7517"*), but the `Array` branch of `Set#initialize` still did `jwks.map { |k| JWT::JWK.new(...) }` with no rescue, so building a set from a bare array of JWKs remained fatal on one unrecognised key.

RFC 7517 §5 says a processor SHOULD ignore key types it does not understand. This is becoming a live issue as post-quantum keys (ML-DSA, `kty:"AKP"`, RFC 9964) start appearing in published key sets: one unknown key otherwise stops the classical keys next to it from working. Relates to #743. (Note: the Hash path is already fine on `main` thanks to #728; this closes the array-input gap.)

### Change

- `Set#initialize` Array branch now uses the same `each_with_object` + `rescue JWT::UnsupportedKeyType` pattern as the Hash branch.
- Added a spec mirroring the existing Hash-case test.

`spec/jwt/jwk/set_spec.rb` passes (20 examples, 0 failures).

### DIFF
--- a/lib/jwt/jwk/set.rb
+++ b/lib/jwt/jwk/set.rb
@@ -28,7 +28,11 @@ module JWT
                     nil
                   end
                 when Array
-                  jwks.map { |k| JWT::JWK.new(k, nil, options) }
+                  jwks.each_with_object([]) do |k, arr|
+                    arr << JWT::JWK.new(k, nil, options)
+                  rescue JWT::UnsupportedKeyType
+                    nil
+                  end
                 else
                   raise ArgumentError, 'Can only create new JWKS from Hash, Array and JWK'
                 end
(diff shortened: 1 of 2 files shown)


anakinj (approved):

anakinj:
The change looks good to me and as mentioned in line with what already happens. Rubocop is not happy for some reason, did not check closer what is going on there. Do we need to refactor the initializer a bit to fit the rules....

anakinj:
@RalphBragg This looks good. I refactored the initializer and it's on `main` now. A rebase and conflict fix later and we should be able to merge this.

